PR #10, opened 2015-05-29
Fix parsing bug when STRLEN is used.

edgarRd:
It seems like the order of the rules for unary functions is important (or any
rules in a list of options for that matter), when
using `STRLEN` in a query, the parser is matching `STR` first.
Rearranging the rules seems to fix it.


RubenVerborgh:
Great work, thanks, will merge soon!

RubenVerborgh:
Thanks again, released as v1.1.1!

edgarRd:
Great! Thanks!

PR #15, opened 2015-11-10
Values Clause vs Inline Data Block

QuarksToQuasars:
SPARQL 1.1 http://www.w3.org/TR/sparql11-query/#grammar  distinguishes between values clauses [28] and inline data blocks [61], with the difference that the values clause can also be used at end of the query or sub-select.

This PR is to add the distinction to the jison grammar. A simple test case for values clause at the end of the query is provided and two existing test cases are being fixed.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -31,8 +31,6 @@ function toQuery(q) {
 
   if (q.updates)
     query += mapJoin(q.updates, toUpdate, ';\n');
-  else if (q.values)
-    query += patterns.values(q);
 
   if (q.group)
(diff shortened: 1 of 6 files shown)


RubenVerborgh:
Thanks, this looks correct indeed. Will investigate in more detail, merge, and release.

RubenVerborgh:
I'm very sorry, this was long overdue but accepted now. While I agree with the fixes of the test cases, and the additional case you've added, I found the same can be achieved without an extra production rule. I've committed this in a353322404ca34120da8bc93961020648fc3d45c. What you think?

PR #18, opened 2016-04-01
Created dist files

Jonas-WMDE:
Just added dist files.


RubenVerborgh:
@Jonas-WMDE Thanks for creating a pull request. However, I don't like to have compiled files in a repository; for me, repositories are only for source code.

Could you explain why you need the dist files? Then we might be able to find an alternative solution (such as a different branch).

Jonas-WMDE:
We don't use browserify in our project that is why I created the dist files. I think you should provide them, so it is very easy for people to use your code. I guess it is very common to provide compiled/min files why you don't like it?

RubenVerborgh:
@Jonas-WMDE I can surely _provide_ the dist files. I also do this for N3.js in the [browser branch](https://github.com/RubenVerborgh/N3.js/tree/browser).

However, I don't think that dist files have a place in the main repository:
- Repositories are for source code, not for compiled code. (Just like Java projects also don't come with the compiled jars.)
- Having such files in the repository means having to recompile them on every commit.
- It leads to terrible conflicts when merging and for pull requests.

Would it be a solution for you to provide the dist files in another branch?

Jonas-WMDE:
If you don't like to have it in the repo maybe you could provide release zip-files. Having it in another branch would also work for me.

RubenVerborgh:
Releases are now available in the [dist branch](https://github.com/RubenVerborgh/SPARQL.js/tree/dist).

JonasKress:
Sorry to bug you again, but I'd like to install sparqljs via npm and as there is no dist folder I can not use this :(

RubenVerborgh:
Via npm, you don't need the dist folder; the module can just be used as-is when you do `npm install sparqljs`. Can you confirm?

JonasKress:
We are using npm to pull packages for our web app and we don't have browserify in our build chain. That is why it is not working for us and most npm packages which are also meant to be used in web apps provide a dist folder with builds in it.

RubenVerborgh:
Mmm, npm being the Node Package Manager, I would say that this is not a good fit for Web apps (browserify is). However, it might indeed make sense to distribute the dist folder with the npm release, just like [we distribute the compiled version of the parser](https://github.com/RubenVerborgh/SPARQL.js/commit/d21422e1269f518e8a9748f1f77b65712d744d08). Tracking this in #28.

PR #25, opened 2016-04-11
This is a hack to demonstrate using prefixes in SPARQL generator

JonasKress:
This is a evil hack polluting global state to demonstrate how prefixes could be used in SPARQL generator.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -2,11 +2,14 @@ var XSD_INTEGER = 'http://www.w3.org/2001/XMLSchema#integer';
 
 module.exports = function SparqlGenerator() { return { stringify: toQuery }; };
 
+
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
That's a start already. The evil hack fails in Node though, where there is no `window` ;-)

To do this properly, we'll need to promote `SparqlGenerator` to a proper "class" with membership properties, i.e., `prefixes` must become a field of `SparqlGenerator`.

Note that the current prefix-adding code is not safe: it will break URIs that contain special characters that are not allowed in prefixes. For a conservative example, see [N3.js](https://github.com/RubenVerborgh/N3.js/blob/v0.4.5/lib/N3Writer.js#L117).

RubenVerborgh:
Closing this for now since not mergeable in current form; updates are most welcome though.

PR #27, opened 2016-05-06
Update README.md highlighting added value

federicomorando:
Update README.md to make clear that a great advantage of your approach is to build SPARQL queries dynamically within the js application (and not just having a "wrapper" for making SPARQL queries as other projects do).


RubenVerborgh:
Thanks for bringing this to my attention, @federicomorando. I have opted for a slightly different fix though, since this library does more than just building SPARQL queries.

PR #29, opened 2016-06-07
correctly handle the full URIs of the literal datatypes

joachimvh:
I admit, I did this instead of a mail just to look cool.

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -141,7 +141,7 @@
 
   // Creates a literal with the given value and type
   function createLiteral(value, type) {
-    return '"' + value + '"^^' + type;
+    return '"' + value + '"^^<' + type + '>';
   }
 
   // Creates a triple with the given subject, predicate, and object


RubenVerborgh:
I prefer pull requests over mails 😄 

However, the [N3.js representation](https://github.com/RubenVerborgh/N3.js#utility) does not use angular brackets for IRIs, also not for datatypes. So SPARQL.js is consistent with that.

joachimvh:
You know, I vaguely seem to recall having this exact same discussion a year ago about the fact that this is wrong in my opinion. :D

RubenVerborgh:
It's not really a matter of right or wrong, it's just a design decision 😄 

I just chose to represent IRIs without angular brackets for simplicity, and to be independent of a particular syntax.

joachimvh:
True, 'wrong' wasn't the correct word to use here. You could even say it was the wrong word! (Sorry :stuck_out_tongue_closed_eyes: ).
Reason for asking btw is because I'm looking at my SPARQL algebra generator again and this caused it to contain invalid URIs. So best solution is to use some form of regex to handle this?

RubenVerborgh:
It depends—what syntax does your SPARQL algebra generator use? If it is Turtle, you should use a Turtle serializer (the one in N3.js would do great).

If it is something else, but uses Turtle-style literals, you will need to do more than just adding `<>`, since some characters will need to be escaped. [See this example.](https://github.com/RubenVerborgh/SPARQL.js/blob/v1.1.4/lib/SparqlGenerator.js#L188-L194)

PR #32, opened 2016-11-14
Preserve FILTER clause position when parsing query.

AlexeyMz:
This PR removes special case for `FILTER` when reducing `GroupGraphPattern`.

For example, without this changes the following query
```
SELECT * WHERE {
  ?s ?p ?o .
  FILTER(?o != "foo")
  OPTIONAL { ?s a ?t }
}
```
will become
```
SELECT * WHERE {
  ?s ?p ?o .
  OPTIONAL { ?s a ?t }
  FILTER(?o != "foo")
}
```
after parse -> stringify.

The problem is [order matters in SPARQL](https://wiki.blazegraph.com/wiki/index.php/SPARQL_Order_Matters). In our case, OPTIONAL <-> JOIN reordering causes significant performance issues because database cannot properly optimize latter query.

### DIFF
--- a/test/SparqlParser-test.js
+++ b/test/SparqlParser-test.js
@@ -37,6 +37,15 @@ describe('A SPARQL parser', function () {
     expect(error.message).to.include('Parse error on line 1');
   });
 
+  it('should preserve BGP and filter pattern order', function () {
+    var parser = new SparqlParser();
+    var query = 'SELECT * { ?s ?p "1" . FILTER(true) . ?s ?p "2"  }';
+    var groups = parser.parse(query).where;
+    expect(groups[0].type).to.equal("bgp");
+    expect(groups[1].type).to.equal("filter");
+    expect(groups[2].type).to.equal("bgp");
+  });
+
   describe('with pre-defined prefixes', function () {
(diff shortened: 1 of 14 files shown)


RubenVerborgh:
Order matters indeed. SPARQL.js should do as few transformations as possible.

Will investigate and merge, thanks!

test/SparqlParser-test.js
```diff
@@ -37,6 +37,15 @@ describe('A SPARQL parser', function () {
     expect(error.message).to.include('Parse error on line 1');
   });
 
+  it('should preserve BGP and filter pattern order', function () {
+    var parser = new SparqlParser();
+    var query = 'SELECT * { ?s ?p "1" . FILTER(true) . ?s ?p "2"  }';
```
RubenVerborgh: This is actually a case where order doesn't matter, but I suppose SPARQL.js shouldn't mess with the order in any case. Query optimizers can always later move things around.

RubenVerborgh:
Published as v1.2.0.

AlexeyMz:
@RubenVerborgh thanks :+1:

PR #33, opened 2016-12-01
Add space before . (dot) when serializing triple

AlexeyMz:
When serializing a query with triple that contains `xsd:integer` in an object position, currently it's being serialized as `n.`

While it's technically correct by the specification, it's still preferable to serialize it as `n .` because there are some broken triplestore parsers.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -66,7 +66,11 @@ function toPattern(pattern) {
 }
 var patterns = {
   triple: function (t) {
-    return toEntity(t.subject) + ' ' + toEntity(t.predicate) + ' ' + toEntity(t.object) + '.';
+    var objectEntity = toEntity(t.object);
+    // prevent syntax error in SPARQL 1.0 parsers when dot after a triple
+    // parsed as decimal separator instead of triple separator
+    if (isDigitSequence(objectEntity)) { objectEntity += ' '; }
+    return toEntity(t.subject) + ' ' + toEntity(t.predicate) + ' ' + objectEntity + '.';
   },
   array: function (items) {
     return mapJoin(items, toPattern, '\n');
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
Matter of taste, I guess… I'm not a fan of the space before dots. Any pointers to broken parsers?

AlexeyMz:
Actually I'm not sure if this is issue with parsers or if it actually is not valid according to standard.

Let's say we have query like:
```
select distinct ?Concept where { ?Concept a 1. ?Concept2 a 1.} LIMIT 100
```
This query produces syntax errors at http://sparql.europeana.eu/ or https://query.wikidata.org/

I think it's related to this BNF rule in the spec:
```
[78]  	DECIMAL	  ::=  	[0-9]+ '.' [0-9]* | '.' [0-9]+
```
From the notes there:
> When tokenizing the input and choosing grammar rules, the longest match is chosen.

RubenVerborgh:
The actual production for [DECIMAL in SPARQL 1.1](https://www.w3.org/TR/sparql11-query/#rDECIMAL) is

```
[147]  		DECIMAL	  ::=  	[0-9]* '.' [0-9]+
```

which differs from [SPARQL 1.0](https://www.w3.org/TR/rdf-sparql-query/#rDECIMAL).

So my guess is that the engines you mention are SPARQL 1.0 engines.

The distinction is important because, if the query is indeed invalid, the SPARQL.js parser should fail on it. However, that doesn't seem to be the case.

Perhaps as a counterproposal, how about we only add a space after integers then? This would bring us SPARQL 1.0 compatibility.

aindlq:
I would say it can be related to the DOUBLE rule:
```
[148]  	DOUBLE	  ::=  	[0-9]+ '.' [0-9]* EXPONENT | '.' ([0-9])+ EXPONENT | ([0-9])+ EXPONENT
```
where it matches `[0-9]+ '.'` and then expects EXPONENT.

RubenVerborgh:
Nah, EXPONENT is harmless:

```
[155]  	EXPONENT	  ::=  	[eE] [+-]? [0-9]+
```

Nothing else is going to start with `E`.

Must really be the SPARQL 1.0/1.1 difference; otherwise, this parser would also fail on the query.

aindlq:
One more example of parser that complains about the query from the above: http://yasqe.yasgui.org/

RubenVerborgh:
Well, looking at the BNF syntax of SPARQL 1.1, those parsers are wrong to reject it.

However, I'd be happy to update the generator to add a space after integers.

RubenVerborgh:
Implemented in 7a19244caf9ac65d5ef3e893c35843703089fa5f.

PR #36, opened 2017-04-14
Use prefixes when generating IRIs in SPARQL, add an option to remove unused prefixes

### LINKED ISSUE #24: When generating SPARQL use prefixes
I think it would be nice to use prefixes when generating entities of a SPARQL query and maybe also delete prefixes from the list that are not actually in use. 

So in case of Wikidata this: 
http://www.wikidata.org/prop/direct/P31 

would be rendered as:
wdt:P31

What do you think?
It could also be an option in the generator.

Cheers,
Jonas

AlexeyMz:
Closes #24.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -1,13 +1,32 @@
 var XSD_INTEGER = 'http://www.w3.org/2001/XMLSchema#integer';
 
-module.exports = function SparqlGenerator() {
+/**
+ * @param options {
+ *   allPrefixes: boolean
+ * }
+ */
+module.exports = function SparqlGenerator(options) {
   return {
-    stringify: function (q) { return new Generator(q.prefixes).toQuery(q); }
+    stringify: function (q) { return new Generator(options, q.prefixes).toQuery(q); }
   };
 };
 
-function Generator(prefixes) {
-  this.prefixes = prefixes || {};
+function Generator(options, prefixes) {
+  this.options = options || {};
+
+  prefixes = prefixes || {};
+  var prefixesArray = Object.keys(prefixes);
+  var iriList = prefixesArray
+    .map(function (prefix) { return prefixes[prefix] })
+    .join('|')
+    .replace(/[\]\/\(\)\*\+\?\.\\\$]/g, '\\$&');
+  this.prefixRegex = new RegExp('^(' + iriList + ')[a-zA-Z][\\-_a-zA-Z0-9]*$', 'g');
+  this.prefixByIri = prefixesArray.reduce(function (acc, prefix) {
+    acc[prefixes[prefix]] = prefix;
+    return acc;
+  }, {});
+  this.usedPrefixes = {};
+
   this.patterns = {};
   for (var key in PATTERNS) {
     this.patterns[key] = PATTERNS[key].bind(this);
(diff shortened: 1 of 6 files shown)


bin/json-to-sparql
```diff
@@ -15,6 +15,6 @@ var query = '', source = args[0] ? fs.createReadStream(args[0]) : process.stdin;
 source.setEncoding('utf8');
 source.on('data', function (data) { query += data; });
 source.on('end', function () {
-  var sparqlText = new SparqlGenerator().stringify(JSON.parse(query));
+  var sparqlText = new SparqlGenerator({removeUnusedPrefixes: true}).stringify(JSON.parse(query));
```
RubenVerborgh: I would make this the default, and rather have `allPrefixes` as an option.

lib/SparqlGenerator.js
```diff
@@ -1,14 +1,31 @@
...
+ */
+module.exports = function SparqlGenerator(options) {
+  return {
+    stringify: function (q) { return new Generator(options, q.prefixes).toQuery(q); }
+  };
+};
+
+function Generator(options, prefixes) {
+  this.options = options || {};
+
```
RubenVerborgh: Superfluous newline.

lib/SparqlGenerator.js
```diff
@@ -226,28 +256,44 @@ var escape = /["\\\t\n\r\b\f]/g,
     escapeReplacements = { '\\': '\\\\', '"': '\\"', '\t': '\\t',
                            '\n': '\\n', '\r': '\\r', '\b': '\\b', '\f': '\\f' };
 
+Generator.prototype.toEntityIRI = function (iri) {
```
RubenVerborgh: The function name is unclear.

lib/SparqlGenerator.js
```diff
@@ -226,28 +256,44 @@ var escape = /["\\\t\n\r\b\f]/g,
...
+  var longestPrefix, prefixLength = Number.NEGATIVE_INFINITY;
```
RubenVerborgh: 0 or -1 should be fine

lib/SparqlGenerator.js
```diff
@@ -226,28 +256,44 @@ var escape = /["\\\t\n\r\b\f]/g,
...
+  for (var prefix in this.prefixes) {
```
RubenVerborgh: Looping every time is expensive. Consider a technique such as https://github.com/RubenVerborgh/N3.js/blob/v0.9.1/lib/N3Writer.js#L225-L226

lib/SparqlGenerator.js
```diff
@@ -257,7 +303,9 @@ function toUpdate(update) {
 function isString(object) { return typeof object === 'string'; }
 
 // Maps the array with the given function, and joins the results using the separator
-function mapJoin(array, func, sep) { return array.map(func).join(isString(sep) ? sep : ' '); }
+Generator.prototype.mapJoin = function (array, func, sep) {
```
RubenVerborgh: This utility function can remain a non-member.

RubenVerborgh:
Please amend to relevant commits for clarity.

lib/SparqlGenerator.js
AlexeyMz: Hmm, then it would be required to either bind `this` to `func` at each call site or provide `this` to `mapJoin` as another argument.

RubenVerborgh: Other argument seems fine. Perhaps the argument order should be `array, sep, func, self`.

RubenVerborgh:
Thanks!

PR #37, opened 2017-04-19
Support inherited prefixes when generating IRIs

AlexeyMz:
This PR adds support for inherited prefixes in a prefixes map for SPARQL generator.


RubenVerborgh:
Good find, thanks!

RubenVerborgh:
Published as v1.4.1.

AlexeyMz:
Thanks!

PR #38, opened 2017-05-04
Disable explicit group pattern collapse

### LINKED ISSUE #35: Disable (minor) query rewrites by default
Let's say we have a simple query like:

```sparql
SELECT * WHERE {
  { ?s ?p ?o. }
  ?a ?b ?c .
}
```
With the last SPARQL.js it is parsed to: 

```json
{
  "queryType": "SELECT",
  "variables": [
    "*"
  ],
  "where": [
    {
      "type": "bgp",
      "triples": [
        {
          "subject": "?s",
          "predicate": "?p",
          "object": "?o"
        },
        {
          "subject": "?a",
          "predicate": "?b",
          "object": "?c"
        }
      ]
    }
  ],
  "type": "query",
  "prefixes": {}
}
```

Which doesn't correspond to the original query, if I serialize it, the result will be: 

```sparql
SELECT * WHERE {
  ?s ?p ?o.
  ?a ?b ?c .
}
```

This missing group sometimes can be significant, e.g. in my use-case I'm trying to apply blazegraph(https://wiki.blazegraph.com/wiki/index.php/QueryHints) query hint to the subgroup.

AlexeyMz:
Closes #35.

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -226,6 +226,32 @@
     });
     return unionAll(objects, otherTriples || [], triples);
   }
+
+  function collapseAjacentBGPs(ajacentPatterns) {
+    // Simplify the groups by merging adjacent BGPs
+    var groups = [], currentBgp;
+    for (var i = 0, group; group = ajacentPatterns[i]; i++) {
+      switch (group.type) {
+        // Add a BGP's triples to the current BGP
+        case 'bgp':
+          if (group.triples.length) {
+            if (!currentBgp)
+              appendTo(groups, currentBgp = group);
+            else
+              appendAllTo(currentBgp.triples, group.triples);
+          }
+          break;
+        // All other groups break up a BGP
+        default:
+          // Only add the group if its pattern is non-empty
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Makes sense, thanks. Let me think about this for a bit, and I'll merge.

RubenVerborgh:
The problem is basically that this would break backward compatibility, so we'd need a new major version. So I'm considering whether this is the way to go, or whether this should be behind a flag until the next major version. It might still be useful to place the algorithm somewhere.

AlexeyMz:
I think it would be fine to keep the changes under a flag to preserve compatibility for now.

RubenVerborgh:
Alright, would you mind adding that?

AlexeyMz:
I added `collapseGroups` flag (enabled by default).

RubenVerborgh:
Thanks, will take this up ASAP!

RubenVerborgh:
Merged as b259983b1d91461cfb91af55750606a7100f8e15.

PR #39, opened 2017-05-15
Fix lost UNION sub-groups when collapseGroups is false

AlexeyMz:
This is a fix for an error, introduced in b259983.

Maybe it makes sense to replace this block with
```
if ($1.length) {
  $$ = { type: 'union', patterns: unionAll($1.map(degroupSingle), [degroupSingle($2)]) };
} else {
  $$ = Parser.options.collapseGroups ? degroupSingle($2) : $2;
}
```
to make it easier to follow the logic here?

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -617,7 +617,8 @@ GraphPatternNotTriples
     : ( GroupGraphPattern 'UNION' )* GroupGraphPattern
     {
       if (!Parser.options.collapseGroups)
-        $$ = !$1.length ? $2 : { type: 'union', patterns: unionAll($1, $2) };
+        $$ = !$1.length ? $2 :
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Yeah, should have tested this instead just changing what you originally wrote :-$

However, I don't immediately understand why `degroupSingle` depends on `Parser.options.collapseGroups` in the second case but not the first.

AlexeyMz:
With current generator it's required to collapse immediate children groups of UNION; otherwise there would be an extra group around UNION contents when serializing a query.

RubenVerborgh:
Okay, makes sense. Sorry for messing up that commit, will merge this ASAP.

RubenVerborgh:
Merged as accb3af21d13cc41a0b812cda1993fc2f3d0d0db.

Issue #44, opened 2017-08-10
Improve stringification of queries


lucaswerkmeister:
On [query.wikidata.org](https://query.wikidata.org/) we use SPARQL.js to edit queries, which includes converting the parsed form back to SPARQL after editing it. This means that if the user starts with a query like

```sparql
SELECT ?person WHERE {
  ?person wdt:P31 wd:Q5;
          wdt:P1412 wd:Q1860, wd:Q188;
          wdt:P22 [ wdt:P27 wd:Q142 ].
}
```

and edits it, the result they get will be

```sparql
SELECT ?person WHERE {
  ?person wdt:P31 wd:Q5.
  ?person wdt:P1412 wd:Q1860.
  ?person wdt:P1412 wd:Q188.
  ?person wdt:P22 _:b1.
  _:b1 wdt:P27 wd:Q142.
}
```

which is significantly less readable.

It would be awesome if the conversion to string would automatically use the following abbreviations when possible:

- use `;` when the subject is repeated
- use `,` when the subject and predicate are repeated
- use `[ ]` when the object is a blank node that is only used as subject of immediately subsequent triples

RubenVerborgh:
Good idea. This should be really simple; the [relevant code from N3.js](https://github.com/RubenVerborgh/N3.js/blob/v0.11.1/lib/N3Writer.js#L74-L84) can nearly be duplicated (taking care of special cases such as property paths etc.).

lucaswerkmeister:
Oh, I didn’t know N3.js did this. But that only takes care of `;` and `,`, not `[]`, if I’m reading it correctly?

RubenVerborgh:
Yes, because N3.js has a streaming serializer; there's no way to predict whether a blank node will be reused. That's different with SPARQL.

PR #45, opened 2017-08-10
WIP: Improve query stringification

lucaswerkmeister:
Abbreviate subsequent triples with the same subject with `;`, and subsequent triples with the same subject and predicate with `,`.

Yes, the tests don’t pass.

---

This is nowhere *near* ready for merging yet, I just wanted to upload what I have so there’s no duplicate work when someone else also wants to start working on #44.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -90,16 +90,63 @@ Generator.prototype.toPattern = function (pattern) {
   return this[type](pattern);
 }
 
-Generator.prototype.triple = function (t) {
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
I think the grouping is not needed/desired: I'd suggest just going through the triples in order, and remembering the previous subject and predicate [as in N3.js](https://github.com/RubenVerborgh/N3.js/blob/v0.11.1/lib/N3Writer.js#L74-L84).

RubenVerborgh:
Closing this in the meantime; feel free to reopen on completion.

PR #48, opened 2017-10-02
Disable group collapse

joachimvh:
Note that the previous PR should be accepted first since this one builds on it (one new test is changed).

This change will make some results more verbose where it might not be needed.
The biggest example of this is test sparql-12.json,
where a new group is added just to contain the subquery.


RubenVerborgh:
Can you remind me why we are removing this option, as opposed to disabling it by default?

joachimvh:
Well the problem was that we kept getting cases where the collapsing introduced incorrect results. E.g. the new test case added in this pull request, `SELECT * { {} OPTIONAL {SELECT * { ?s ?p ?o }} }`, resulted in the following output:
```json
{
  "queryType": "SELECT",
  "variables": [
    "*"
  ],
  "where": [
    {
      "queryType": "SELECT",
      "variables": [
        "*"
      ],
      "where": [
        {
          "type": "bgp",
          "triples": [
            {
              "subject": "?s",
              "predicate": "?p",
              "object": "?o"
            }
          ]
        }
      ],
      "type": "optional"
    }
  ],
  "type": "query",
  "prefixes": {}
}
```
Notice that we have something of type optional here with a where clause. Since this was not the first case, the question was if there might still be other errors introduced by the collapsing.

Another example is the following 2 queries that produce the same output even though they have a different interpretation:
```
PREFIX : <http://www.example.org>
SELECT *
WHERE {
    :s :p ?o .
    :s :q ?o1
    { BIND((1+?o) AS ?o1) }
}
```
```
PREFIX : <http://www.example.org>
SELECT *
WHERE {
    :s :p ?o .
    :s :q ?o1
    BIND((1+?o) AS ?o1)
}
```

RubenVerborgh:
Makes sense, thanks! The disappearing option makes this a breaking version change though, so will release as a new major version.

PR #50, opened 2018-01-11
Remove invalid hidden character

rubensworks:
sparql.js contains a hidden character.
Node.js seems to be able to handle it, but when packaged using webpack, Chrome errors on it.
This commit fixes that by removing the hidden character.


RubenVerborgh:
You're probably not serving it as UTF-8 :wink:

PR #52, opened 2018-01-11
Support GROUP BY variables when generating SPARQL

joachimvh:
Currently the variable in a GROUP BY element is ignored when generating the corresponding SPARQL string.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -48,7 +48,8 @@ Generator.prototype.toQuery = function (q) {
 
   if (q.group)
     query += 'GROUP BY ' + mapJoin(q.group, undefined, function (it) {
-      return isString(it.expression) ? it.expression : '(' + this.toExpression(it.expression) + ')';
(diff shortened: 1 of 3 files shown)


lib/SparqlGenerator.js
```diff
@@ -48,7 +48,8 @@ Generator.prototype.toQuery = function (q) {
 
   if (q.group)
     query += 'GROUP BY ' + mapJoin(q.group, undefined, function (it) {
-      return isString(it.expression) ? it.expression : '(' + this.toExpression(it.expression) + ')';
+      var result = isString(it.expression) ? it.expression : '(' + this.toExpression(it.expression) + ')';
+      return it.variable ? '(' + result + ' AS ' + it.variable + ')' : result;
```
RubenVerborgh: Does this mean we'll always generate the `AS` now, even when we don't need it?

joachimvh: No, the check on the second line makes sure it only gets added when the variable value is defined on the object. (Else all the other group by tests would fail).

PR #53, opened 2018-02-08
Place brackets around order expressions

joachimvh:
Currently, the generator would generate the following query:
```sparql
PREFIX : <http://example.org/ns#>
SELECT * { ?s ?p ?o }
ORDER BY ?o+5
```
Which is invalid and can't be parsed again by the parser due to the missing brackets around `?o+5`.

Although it is not necessary to place brackets around all expressions, it is also not wrong. So this felt like the more cleaner solution for me instead of adding exceptions for specific cases.


RubenVerborgh:
You're right, I also favor being on the safe side. Thanks!

PR #54, opened 2018-02-08
Place parentheses for path postfix operators

joachimvh:
This is to prevent problems when having nested path operators.

```sparql
prefix : <http://example.org/>
select ?X where { :A0 ((:P)*)* ?X }
order by ?X
```
When parsing this query and converting it back to sparql the result was the following:
```
prefix : <http://example.org/>
select ?X where { :A0 :P** ?X }
order by ?X
```
Which is invalid due to the double path operators.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -251,7 +251,7 @@ Generator.prototype.toEntity = function (value) {
     case '*':
     case '+':
     case '?':
-      return items[0] + path;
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Oooh lowercase sparql looks ugly. Thanks nonetheless :wink:

joachimvh:
Blame these guys ;) https://www.w3.org/2009/sparql/docs/tests/

PR #55, opened 2018-03-27
remove irregular white spaces

e-e-e:
When using browserify on a module that requires SPARQL.js I get errors because of irregular whitespaces being transformed incorrectly. I know that this is actually an issue with browserify, but thought that it would be useful also to clean up the white spaces here anyway - as they can cause other issues too.

Perhaps its worth adding eslint to this library with the no-irregular-whitespace rule.

Also - thanks for making an awesome sparql parser!

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -427,7 +427,7 @@ PN_LOCAL_ESC          "\\"("_"|"~"|"."|"-"|"!"|"$"|"&"|"'"|"("|")"|"*"|"+"|","|"
 QueryOrUpdate
     : Prologue ( Query | Update? ) EOF
     {
-      $2 = $2 || {};
+      $2 = $2 || {};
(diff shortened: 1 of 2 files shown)


e-e-e:
Further investigation suggests that its not browerify's fault at all, but is actually an issue when irregular white space is read by the browser - the browser does not read them as white space.

RubenVerborgh:
> the browser does not read them as white space.

It does, if the JavaScript files are served correctly as UTF-8. But to also make the library work in case of misconfiguration, this can help.

e-e-e:
No problem - I will amend that. Apologies for my lack of discernment - just a zealous find and replace on my part.

e-e-e:
@RubenVerborgh I removed the changes to the .md files.

RubenVerborgh:
Thanks a lot!

PR #57, opened 2018-06-04
Fix tests on Windows, use locally installed tools in NPM scripts

AlexeyMz:


package.json
```diff
@@ -20,8 +20,8 @@
   "main": "sparql.js",
   "bin": "bin/sparql-to-json",
   "scripts": {
-    "build": "jison lib/sparql.jison -p slr -o lib/SparqlParser.js",
-    "test": "mocha"
+    "build": "./node_modules/.bin/jison lib/sparql.jison -p slr -o lib/SparqlParser.js",
+    "test": "./node_modules/.bin/mocha"
```
RubenVerborgh: This shouldn't be necessary; npm should take care of that.
@joachimvh As a Windows user, can you comment?

test/SparqlGenerator-test.js
```diff
@@ -38,7 +39,8 @@ describe('A SPARQL generator', function () {
 
     it('should remove unused prefixes from "' + query + '"', function () {
       var parsedQuery = parseJSON(fs.readFileSync(parsedQueryFile, 'utf8'));
-      var expectedQuery = fs.readFileSync(generatedQueryFile, 'utf8');
+      var expectedQuery = fs.readFileSync(generatedQueryFile, 'utf8')
+        .split(os.EOL).join('\n');
```
RubenVerborgh: I think we can just `.replace(/[\n\r]+/g, '\n')`

package.json
joachimvh: `npm run-script build` and `npm test` work for me from the command line without having to change anything. I do have `mocha` installed as global dependency (but not `jison` so that shouldn't make a difference)

RubenVerborgh: @AlexeyMz Any insight in why it wouldn't work on your system?

test/SparqlGenerator-test.js
RubenVerborgh: @joachimvh You running into this problem as well?

joachimvh: No but I do have my git setup to use LF instead of CRLF line endings. I did just try manually storing some of the files with CRLF line endings now and the tests still work, so not sure what the problem is.

RubenVerborgh:
Closing this for now given @joachimvh's input, but feel free to reopen if needed.

test/SparqlGenerator-test.js
AlexeyMz: @RubenVerborgh `.replace(/[\n\r]+/g, '\n')` will remove all empty lines. I use default Git setup for multiplatform projects on Windows, so it converts `\n` to `\r\n` on checkout; see `core.autocrlf` section here: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

RubenVerborgh:
Re-opening for line-breaking issue.

test/SparqlGenerator-test.js
RubenVerborgh: Right, and we have empty lines after prefixes.

How about `.replace(/\n\r/g, '\n')` then? Feel free to amend and I'll merge.

PR #58, opened 2018-07-16
Stringify BGP triples with repeated subject and predicate using semicolon (;) and comma (,)

### LINKED ISSUE #44: Improve stringification of queries

AlexeyMz:
Fixes #44, except encoding blank nodes using `[ ]` (this requires to traverse the whole query to determine whether it's safe to replace blank node reference with abbreviated form as block).

I also included the fix from PR #57 (fix failing tests on Windows due to different new line characters).


RubenVerborgh:
Woops, merged #57, can you rebase?

AlexeyMz:
Done.

PR #59, opened 2018-07-16
Support CLEAR and DROP operations in SparqlGenerator

AlexeyMz:
From spec:
https://www.w3.org/TR/sparql11-update/#clear
https://www.w3.org/TR/sparql11-update/#drop

PR #60, opened 2018-08-07
Allow for a semi-colon before the final period in CONSTRUCT queries

shad:
To match the spec...

See: https://www.w3.org/TR/sparql11-query/#rPropertyListNotEmpty

This indicates that the following _should_ be legal...

```
CONSTRUCT {
  ?s ?p ?o; 
     a <http://example.org/Thing>; 
  .
} WHERE {
  ?s ?p ?o
}
```
But it doesn't currently parse correctly. We see this pattern fairly frequently in our app. This is valid in Turtle, so people assume it's valid in `CONSTRUCT` queries as well.

NOTE: This is my first attempt at working with a JISON grammar like this, so please tell me if there's a better way to accomplish what I'm trying to (or if this doesn't make sense).

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -654,7 +654,7 @@ PropertyList
     : PropertyListNotEmpty?
     ;
 PropertyListNotEmpty
-    : ( VerbObjectList ';'+ )* VerbObjectList -> unionAll($1, [$2])
+    : ( VerbObjectList ';'+ )* VerbObjectList ';'? -> unionAll($1, [$2])
     ;
 VerbObjectList
     : Verb ObjectList -> objectListToTriples($1, $2)
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -654,7 +654,10 @@ PropertyList
     : PropertyListNotEmpty?
     ;
 PropertyListNotEmpty
-    : ( VerbObjectList ';'+ )* VerbObjectList ';'? -> unionAll($1, [$2])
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
Thanks! In Turtle, there can even be multiple semicolons, but SPARQL doesn't seem to support that. @joachimvh Can you verify and merge?

joachimvh:
Actually you can have multiple semicolons according to the spec linked above. Didn't actually know that :D
```
[77]  PropertyListNotEmpty  ::=  Verb ObjectList ( ';' ( Verb ObjectList )? )* 
```

I'm also not that great at jison, but from what I tested out, it can be simplified to the following

```
PropertyListNotEmpty
    : VerbObjectList ( SemiOptionalVerbObjectList )* -> unionAll([$1], $2)
    ;
SemiOptionalVerbObjectList
    : ';' VerbObjectList? -> unionAll($2)
    ;
```

Could you change it to that?
(From how I understand it this is because there will always be a `SemiOptionalVerbObjectList`, although it might be empty because it contents were optional)

RubenVerborgh:
Right, we might want a separate test for multiple semicolons as well in that case.

shad:
Wow! Thanks for the feedback!  You never know when you submit a patch if you'll ever hear back or not :) I'll make changes and add a test this week.

joachimvh:
Thanks for this!

shad:
Awesome! Thank you! :)

shad:
Question: how often do you cut new NPM versions? (no rush, just planning)

RubenVerborgh:
Depends; I'm just waiting for #58 and #59, and will then release.

PR #62, opened 2018-09-01
Make newline and indention characters configurable

white-gecko:
With this pull request it is possible to set the indention and newline characters through the options parameter. This allows also to put a query in a single line.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -1,16 +1,14 @@
 var XSD_INTEGER = 'http://www.w3.org/2001/XMLSchema#integer';
-var INDENTATION = '  ';
-var NEWLINE = '\n';
 
 function Generator(options, prefixes) {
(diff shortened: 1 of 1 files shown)


white-gecko:
Thank you for the comments.
Sorry for the missing semicolons ;-)
I've changed the usage of the option now and made it the other way round. Is this as you like it?
The only thing that bothers me is, that I had to make the helper function `indent` a method of the `Generator` to gain access to the options. Alternatively we could pass the option to the function. Which solution do you prefer?

RubenVerborgh:
> I had to make the helper function indent a method of the Generator

Perfect, member functions are the way to go. Will check and merge.

RubenVerborgh:
I added 4e0672ceb7cd7b0819066144758ae64ad06d0db1, can you confirm this is okay? I renamed the option to `indent`.

Issue #73, opened 2018-12-18
Switch to RDF/JS representation


iddan:
Currently RDFTerms (Literals, IRIs and prefixed IRIs) are represented as strings.
I suggest to use the RDF/JSON representation for RDF values.
Also, prefixed IRIs should be distinguished from IRIs.
```javascript
import { Parser } from 'sparqljs';
const parser = new Parser();
parser.parse(`
PREFIX foaf:  <http://xmlns.com/foaf/0.1/>
SELECT *
WHERE {
    ?person foaf:name ?name .
    ?person foaf:mbox ?email .
}
`)
```
Should return
```json
{
  "queryType": "SELECT",
  "variables": [
    "*"
  ],
  "where": [
    {
      "type": "bgp",
      "triples": [
        {
          "subject": {
              "type": "variable",
              "identifier": "person"
          },
          "predicate": {
              "type": "prefixed-iri",
              "prefix": "foaf",
              "value": "name"
          },
          "object": {
              "type": "variable",
              "identifier": "name"
          }
        },
        {
          "subject": {
              "type": "variable",
              "identifier": "person"
          },
          "predicate": {
              "type": "iri",
              "value": "http://xmlns.com/foaf/0.1/mbox"
          },
          "object": {
              "type": "variable",
              "identifier": "email"
          }
        }
      ]
    }
  ],
  "type": "query",
  "prefixes": {
    "foaf": "http://xmlns.com/foaf/0.1/"
  }
}
```
We'd be happy to offer a PR

RubenVerborgh:
Thanks for volunteering.

I would propose the RDF/JS interface:
http://rdf.js.org

What you could do is pass in a factory to the parser constructor (RDF/JS
has this). That way, you can have your favorite JSON output, and others can
have their own.

Essentially, the needed changes should just be passing in the factory and
calling its methods to create terms and triples.

Good plan? If so, I can give you some suggestions on how to get started.

Ruben

iddan:
Amazing one! Please do

iddan:
Something I miss in it is an explicit representation for a Prefixed URI. Do you know if it includes it?

RubenVerborgh:
It does not, becaude that is only syntactical, not in RDF terms. But I have
a solution if you need it.

iddan:
I think we should use the RDF/JSON as you suggested and additionally define a prefixed URI and variable types. What do you think?

akuckartz:
JSON-LD ?

[...]

PR #80, opened 2019-03-09
Output RDF/JS terms

namedgraph:
@RubenVerborgh my partial attempt at #73. Is it something along the lines?

I'm not familiar with the setup, just stabbing in the dark here really. So I won't be able to fix the tests easily.

Also not sure if all of `Literal` grammar is covered (where `createLiteral()` is not called).
`XSD_TRUE`/`XSD_FALSE` constants might not be required anymore.

But otherwise I ran some queries through `sparql-to-json` and I'm getting what I think is expected output.


RubenVerborgh:
> @RubenVerborgh my partial attempt at #73. Is it something along the lines?

Looks good indeed!

> I'm not familiar with the setup, just stabbing in the dark here really.

```
npm run build && npm run test
```

For the tests, you might want to inject a factory that gives the old output back; then you don't have to adjust any of the tests.

namedgraph:
Updated. The approach mostly works, but there's a problem with literals, more specifically `TestDataFactory.literal()` (38 tests fail).

`createLiteral()` in JISON attempts to create a `NamedNode` for the datatype, but `TestDataFactory.namedNode()` returns string, and then when `TestDataFactory.literal()` is called it thinks that string is a lang tag and not a datatype.

Some hack is required there to distinguish between lang and datatype literals.

RubenVerborgh:
> then when `TestDataFactory.literal()` is called it thinks that string is a lang tag and not a datatype.

So can we change `TestDataFactory.literal` such that it recognizes things that are not lang tagss?

namedgraph:
I guess first we need to have 2 methods `createdTypedLiteral()`/`createLangLiteral()` in JISON. Because this rule

    String LANGTAG  -> $1 + lowercase($2)

also need wrapping into RDF/JS structure, correct? And current `createLiteral()` won't work because it only deals with typed literals.
I haven' done this yet.

RubenVerborgh:
Can't we just delegate to `Parser.factory.literal`? That will do the right thing, given the right argument types. We'd only have to cheat a little bit for the tests, but that's okay.

Thanks for all the work so far!

lib/sparql.jison
```diff
@@ -13,9 +13,7 @@
       XSD_INTEGER  = XSD + 'integer',
       XSD_DECIMAL  = XSD + 'decimal',
       XSD_DOUBLE   = XSD + 'double',
-      XSD_BOOLEAN  = XSD + 'boolean',
-      XSD_TRUE =  '"true"^^'  + XSD_BOOLEAN,
-      XSD_FALSE = '"false"^^' + XSD_BOOLEAN;
```
RubenVerborgh: You might want to make all of these already `NamedNode`s.

lib/sparql.jison
```diff
@@ -815,14 +812,14 @@ Literal
     | DOUBLE  -> createLiteral(lowercase($1), XSD_DOUBLE)
     | NumericLiteralPositive
     | NumericLiteralNegative
-    | 'true'  -> XSD_TRUE
-    | 'false' -> XSD_FALSE
+    | 'true'  -> createLiteral('true', XSD_BOOLEAN)
+    | 'false' -> createLiteral('false', XSD_BOOLEAN)
```
RubenVerborgh: No need to create new literals every time; we can just reuse constants.

test/TestDataFactory.js
```diff
@@ -0,0 +1,68 @@
...
+    // Convert an integer or double
+    case 'number':
+      if (Number.isFinite(value))
+        datatype = Number.isInteger(value) ? xsd.integer : xsd.double;
+      else {
+        datatype = xsd.double;
+        if (!Number.isNaN(value))
+          value = value > 0 ? 'INF' : '-INF';
+      }
+      break;
```
RubenVerborgh: We probably shouldn't be too smart here. RDF/JS doesn't do conversion either, so it's up to the parser to call this with a string.

test/TestDataFactory.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// ## Module exports
+module.exports = TestDataFactory = {
+  // ### Public factory functions
+  namedNode: namedNode,
+  blankNode: blankNode,
+  variable:  variable,
+  literal:   literal
+};
```
RubenVerborgh: Newline 😉

lib/sparql.jison
namedgraph: But `createLiteral()` second arg is string, not `Term`. I don't think we should treat these specially.

test/TestDataFactory.js
namedgraph: Yeah this part needs to at looked after the datatype/lang issue is resolved.
Probably just removed, I copy-pasted it from N3.js.

namedgraph:
I've updated one more time, hopefully now it's easier to see the issue.

See how now there are 2 methods in JISON for literals: `createLangLiteral()` and `createTypedLiteral()`.

The core issue is that both of them call `Parser.factory.literal()` with string as the 2nd argument:
* `createLangLiteral()` with a "real" string, the lang tag
* `createTypedLiteral()` with a string which is the result of `NamedNode` becoming a string in the `TestDataFactory`, but really means a datatype

Does that make sense? I don't see a very straightforward fix here.

lib/sparql.jison
RubenVerborgh: > But createLiteral() second arg is string

That's a very good reason to make that second arg a named node, because it avoids re-creating all those XSD nodes again and again.

RubenVerborgh:
`test/TestDataFactory.js`:
```
function literal(value, tag) {
  if (tag)
    tag = (tag.length < 10 ? '@' : '^^') + tag;
  return '"' + value + '"' + tag;
}
```

lib/sparql.jison
```diff
@@ -89,8 +87,7 @@
   function toVar(variable) {
     if (variable) {
       var first = variable[0];
-      if (first === '?') return variable;
-      if (first === '$') return '?' + variable.substr(1);
+      if (first === '?' || first === '$') return Parser.factory.variable(variable.substr(1));
```
RubenVerborgh: condition will always be true

namedgraph: I thought so too, but the code disagreed :) I can come back to check this later

test/TestDataFactory.js
```diff
@@ -0,0 +1,36 @@
...
+
+// ### Creates a blank node
+function blankNode(name) {
+  return '_:b' + name;
+}
+
+// ### Creates a literal
+function literal(value, tag) {
+  if (tag) {
+    if (tag.length < 10) // hacky "heuristic" for test purposes only
```
namedgraph: This version of of `literal()` brought the number of test failures down to 10.

But there's something going on with the value escaping, that's why the output differs for lang-tagged and datatyped literals. Try to change any of those and the failure number will go up.

I'm guessing the grammar produces differently escaped literals in different places.

RubenVerborgh:
👍 let me know if you’re stuck, then I can take it from there

namedgraph:
> 👍 let me know if you’re stuck, then I can take it from there

I'm pretty much stuck now :)

test/TestDataFactory.js
```diff
@@ -0,0 +1,36 @@
...
+      return value + '@' + tag; // language tag
+    return value + '^^' + tag; // typed literal
```
namedgraph: I did one more update and removed the quoting from typed literals here.

The failures went up to 38, but all of them involve `xsd:integer` values, so I think it's `SparqlGenerator` treating integers specially, and there is some symmetry missing.

namedgraph:
After a few extra changes all 349 tests pass for me locally. But Travis still fails for some reason.
Please take over now :)

RubenVerborgh:
Great job, thanks. Seems like you need n3 in package.json 😉

RubenVerborgh:
1.0 or up

namedgraph:
And what about `package-lock.json`, should I commit that? Quite a few changed have been generated in it.

test/TestDataFactory.js
```diff
@@ -0,0 +1,38 @@
...
+
+// ### Creates a blank node
+function blankNode(name) {
+  return '_:b' + name;
+}
+
+// ### Creates a literal
+function literal(value, tag) {
+  if (tag) {
+    if (tag.length < 10) // hacky "heuristic" for test purposes only
```
namedgraph: This will break down with short datatype URIs of course :) But good enough to make the tests pass.

namedgraph:
@RubenVerborgh so what does this still need in order to be merged?

namedgraph:
The TypeScript bindings probably need updating too?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sparqljs/index.d.ts

RubenVerborgh:
> @RubenVerborgh so what does this still need in order to be merged?

An hour or two on my side to check everything; planing that this week. Thanks for the work!

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "2.2.1",
   "description": "A parser for the SPARQL query language",
   "author": "Ruben Verborgh <ruben.verborgh@gmail.com>",
+  "dependencies": {
+    "n3": "1.x"
```
RubenVerborgh: `^1.0.4`

sparql.js
```diff
@@ -19,6 +20,7 @@ module.exports = {
     parser.parse = function () {
       Parser.base = baseIRI || '';
       Parser.prefixes = Object.create(prefixesCopy);
+      Parser.factory = factory || N3.DataFactory ;
```
RubenVerborgh: extra space

test/TestDataFactory.js
```diff
@@ -0,0 +1,38 @@
...
+      return value + '@' + tag; // language tag
+    if (tag === "http://www.w3.org/2001/XMLSchema#integer")
+      return '"' + value + '"' + '^^' + tag; // treat integers specially as does SparqlGenerator
+    return value + '^^' + tag; // typed literal
+  }
```
RubenVerborgh: The above is not correct. It should be:
```
function literal(value, tag) {
  var suffix = '';
  if (tag)
    suffix = (/:/.test(tag) ? '^^' : '@') + tag
  return '"' + value + '"' + suffix;
}
```

That's how you go from value/tag to an old N3.js in-memory literal. The trick with the integers was actually masking a bug, which is that the last line of `unescapeString` should be `return string;`.

This fix then leads to a couple of bugs in `SparqlGenerator`, because that class has not yet been updated to RDF/JS.

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "2.2.3",
   "description": "A parser for the SPARQL query language",
   "author": "Ruben Verborgh <ruben.verborgh@gmail.com>",
+  "dependencies": {
+    "n3": "^1.0.4"
```
awwright: Why the need for an N3/Turtle parser? Can this be added as a devDependency?

namedgraph: Good question. @RubenVerborgh maybe https://github.com/rdfjs/data-model would suffice?

RubenVerborgh:
data-model is fine indeed

namedgraph:
So this is not moving anywhere? :)

RubenVerborgh:
Overloaded. Open-source work will likely have to wait until summer.

RubenVerborgh:
Oh wait, I thought this PR only required my review and merge.

But the `SparqlGenerator` code still needs porting. I don't have the bandwidth for that currently.

@rubensworks Is there a student that could look into this? I don't expect this to take more than a day.

rubensworks:
That should be possible, I'll add it to the list.

lib/SparqlGenerator.js
```diff
@@ -1,5 +1,7 @@
 var XSD_INTEGER = 'http://www.w3.org/2001/XMLSchema#integer';
 
+var termToString = require("rdf-string").termToString;
```
rubensworks: Should not be needed, as only variables are stringified. This should be sufficient: `'?' + term.value`

lib/SparqlGenerator.js
```diff
@@ -31,11 +33,12 @@ Generator.prototype.toQuery = function (q) {
   if (q.distinct)
     query += 'DISTINCT ';
 
-  if (q.variables)
+  if (q.variables){
     query += mapJoin(q.variables, undefined, function (variable) {
-      return isString(variable) ? this.toEntity(variable) :
-             '(' + this.toExpression(variable.expression) + ' AS ' + variable.variable + ')';
+      return isString(variable) || isTerm(variable) ? this.toEntity(variable) :
```
rubensworks: @RubenVerborgh Do you want everything to be backwards-compatible with the string representation? Or can everything become terms?

@simonvbrae If so, you can remove all the `isString` stuff, and assume variables are always terms.

lib/sparql.jison
```diff
@@ -810,7 +812,7 @@ GroupConcatSeparator
 Literal
     : String -> createTypedLiteral($1)
     | String LANGTAG  -> createLangLiteral($1, lowercase($2.substr(1)))
-    | String '^^' iri -> createTypedLiteral($1, $3)
+    | String '^^' iri -> createTypedLiteral($1, $3.value)
```
rubensworks: This shouldn't be necessary. I'd change the implementation of `createTypedLiteral` directly to accept a term as second argument.

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "A parser for the SPARQL query language",
   "author": "Ruben Verborgh <ruben.verborgh@gmail.com>",
   "dependencies": {
-    "n3": "^1.0.4"
+    "expect": "^24.8.0",
```
rubensworks: Is this needed? Chai should be sufficient I think.

package.json
```diff
@@ -4,7 +4,9 @@
...
+    "n3": "^1.0.4",
+    "rdf-string": "^1.3.1"
```
rubensworks: Can probably be removed.

test/SparqlGenerator-test.js
```diff
@@ -24,9 +25,13 @@ describe('A SPARQL generator', function () {
...
 
     it('should correctly generate query "' + query + '"', function () {
-      var parsedQuery = parseJSON(fs.readFileSync(parsedQueryFile, 'utf8'));
+      // In parsed query, replace "generated" prefixes with "existing" prefix because all blanknodes are named in the
+      // generated query.
+      var parsedQuery = parseJSON(fs.readFileSync(parsedQueryFile, 'utf8').replace(/g_/g, 'e_'));
       var genQuery = allPrefixesGenerator.stringify(parsedQuery);
-      expect(new SparqlParser(null, null, factory).parse(genQuery)).to.deep.equal(parsedQuery);
+
+      const parsed = new SparqlParser(null, null, null).parse(genQuery);
```
rubensworks: `new SparqlParser()`
Null args should not be needed. (Also later on)

test/SparqlParser-test.js
```diff
@@ -1,15 +1,18 @@
 var SparqlParser = require('../sparql').Parser;
 
-var fs = require('fs'),
-    expect = require('chai').expect;
-
-var factory = require('./TestDataFactory');
+var fs = require('fs');
+    // expect = require('chai').expect;
```
rubensworks: Avoid leaving in commented code.

test/SparqlParser-test.js
```diff
@@ -1,15 +1,18 @@
...
+expect.extend({
+  toEqualParsedQuery,
+});
 
 var queriesPath = __dirname + '/../queries/';
 var parsedQueriesPath = __dirname + '/../test/parsedQueries/';
 
 describe('A SPARQL parser', function () {
-  var parser = new SparqlParser(null, null, factory);
+  var parser = new SparqlParser(null, null, null);
```
rubensworks: I assume this is a mistake? Should the factory not be passed explicitly? (Also later on)

test/SparqlParser-test.js
```diff
@@ -25,7 +28,10 @@ describe('A SPARQL parser', function () {
     it('should correctly parse query "' + query + '"', function () {
       var parsedQuery = parseJSON(fs.readFileSync(parsedQueryFile, 'utf8'));
       query = fs.readFileSync(queriesPath + query + '.sparql', 'utf8');
-      expect(parser.parse(query)).to.deep.equal(parsedQuery);
+
+      const parsed = parser.parse(query);
+      // fs.writeFileSync(parsedQueryFile, JSON.stringify(parsed, null, "  "));
```
rubensworks: Commented code

test/SparqlParser-test.js
```diff
@@ -34,56 +40,62 @@ describe('A SPARQL parser', function () {
...
   describe('with pre-defined prefixes', function () {
     var prefixes = { a: 'ex:abc#', b: 'ex:def#' };
-    var parser = new SparqlParser(prefixes, null, factory);
+    var parser = new SparqlParser(prefixes, null, null);
 
     it('should use those prefixes', function () {
       var query = 'SELECT * { a:a b:b "" }';
-      expect(parser.parse(query).where[0].triples[0])
-        .to.deep.equal({subject: 'ex:abc#a', predicate: 'ex:def#b', object: '""'});
+      var result_json = '{"subject":{"termType":"NamedNode","value":"ex:abc#a"},"predicate":{"termType":"NamedNode","value":"ex:def#b"},"object":{"termType":"Literal","value":"","language":"","datatype":{"termType":"NamedNode","value":"http://www.w3.org/2001/XMLSchema#string"}}}';
```
rubensworks: I suggest skipping the JSON parsing, and just entering this as JS objects directory. (Also later on)

lib/SparqlGenerator.js
RubenVerborgh: Well, there is the case where variables are used as keys to an object, but I don't know if that's what's happening here.

lib/SparqlGenerator.js
simonvbrae: One of the functions using termToString is also called with NamedNode instances. Is this enough reason to keep the module in?

rubensworks: Probably not, as termToString will emit those as `"url"`, while SPARQL requires `<url>`.

simonvbrae: It gets called with BlankNode instances as well so I would suggest to keep it. What do you think?

rubensworks: Bnodes are quite simple as well, that requires just prefixing the value with `"_:"`.

package.json
simonvbrae: This library makes it easier to add your own matcher functions, and since we only used Chai for the expect functionality I thought it made sense to import that functionality seperately.

rubensworks: I'll let @RubenVerborgh decide on this one :-)

lib/sparql.jison
simonvbrae: That is a cleaner solution, thanks for the suggestion.

test/SparqlParser-test.js
simonvbrae: By passing null, I let the parser use it's default factory. That's the factory I would want to pass as the argument anyway, so passing null is just to save us an import.

rubensworks: Ok, in that case this line can be simplified to `new SparqlParser()`.

namedgraph: @RubenVerborgh I think there was a reason why we wanted to pass `TestDataFactory` here?

RubenVerborgh:
That was a temporary measure; should now work without.

lib/SparqlGenerator.js
simonvbrae: I have replaced the check with `=== '*'` so the check is a bit more strict but that's the best we can do since `'*'` is a valid variable at the moment.

package.json
AlexeyMz: I suggest to move this to `devDependencies` as it's only for tests, right?

RubenVerborgh:
@simonvbrae Are there any open questions left?

When resolved, would publish this as `v3.0.0-beta.0` so https://github.com/joachimvh/SPARQLAlgebra.js/pull/49 has some time for testing; once all goes fine, we'll make it a `v3.0.0`.

rubensworks:
I think this is ready for review.

RubenVerborgh:
@rubensworks Are you happy with the `"*"` resolution?
I think I'd have rather seen an object (possibly singleton object), but I can live with it if you can.

rubensworks:
Ah yes, I agree with going for the object solution. Currently the code contains quite a bit of checks such as `if (expr === "*" || isTerm(expr))`, which could be simplified if `*` was a superclass of Term as was discussed above.

@simonvbrae Could you look into this?

simonvbrae:
Sure can, I'll get this done asap.

AlexeyMz:
> I think I'd have rather seen an object (possibly singleton object), but I can live with it if you can.

> Ah yes, I agree with going for the object solution. Currently the code contains quite a bit of checks such as if (expr === "*" || isTerm(expr)), which could be simplified if * was a superclass of Term as was discussed above.

I would strongly suggest to reconsider this for several reasons:
1. This would needlessly break backwards compatibility which means many users will stuck on previous version.
2. Switching to RDFJS by itself breaks ability to trivially serialize query AST to JSON, but it's easy to workaround. Introducing another special term makes it even harder to keep this working.
3. Object identities (including `instanceof` checks) does not work across JS realms, e.g. it's really cumbersome to check if an object has specific type if you pass objects to another browser frame, send it to Worker/MessageChannel or just have multiple library versions at the same time. Basically this is the reason why it's recommended to call `Array.isArray(arr)` instead of `arr instanceof Array`.

RubenVerborgh:
> * This would needlessly break backwards compatibility which means many users will stuck on previous version.

We are breaking backward compatibility in any case; this is a semver major change.

> * Switching to RDFJS by itself breaks ability to trivially serialize query AST to JSON, but it's easy to workaround. Introducing another special term makes it even harder to keep this working.

It would behave the same way as a term, so serialization to JSON would be thee same.

> * Object identities (including `instanceof` checks) does not work across JS realms

Agreed, so there should not be a reliance on object identity; but rather on `x.termType === 'x'` or `x.isWildcard === true` or similar.

AlexeyMz:
@rubensworks I agree, it should be OK to use an object non-identity comparison.

For consistency with other nodes it may look as simple as this (in TypeScript syntax):
```
interface WildcardProjection {
  type: 'wildcard';
}

interface SelectQuery extends BaseQuery {
    queryType: 'SELECT';
    variables: Variable[] | WildcardProjection;
    // ...
}
```

RubenVerborgh:
👍 but with `termType` rather than `type`

AlexeyMz:
May I also suggest to add `type` constants to other node types as well?

Here are the current typings: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d2d3a685fe949a4e938fe2913597e49d1e703ef5/types/sparqljs/index.d.ts

It would be great to make these changes:
1. add `type` equal to `grouping` for Grouping, `ordering` for `Ordering`, `variable` for `VariableExpression`;
2. change `updateType` property to `type` in `InsertDeleteOperation` (currently half operations use `updateType` and the other one just `type`).

RubenVerborgh:
@AlexeyMz Tracking in https://github.com/RubenVerborgh/SPARQL.js/issues/88.

I now see that `type` was coming from sparqljs nodes, not from RDF/JS terms. Wildcard probably can use both.

simonvbrae:
@RubenVerborgh What does this mean for me? Currently Wildcard has termType only

RubenVerborgh:
@simonvbrae `termType` is fine; rest will be followed up in #88.

RubenVerborgh:
@simonvbrae All good from your side?

simonvbrae:
@RubenVerborgh Wildcard support has been added, as far as I'm concerned this is ready for review and possibly to merge.

package.json
RubenVerborgh: Not `devDependency`; it's the default factory.

RubenVerborgh: (We can remove this later; main point is to get to RDF/JS now. Feel free to add an issue.)

lib/SparqlGenerator.js
```diff
@@ -255,30 +256,34 @@ Generator.prototype.toExpression = function (expr) {
 
 // Converts the parsed entity (or property path) into a SPARQL entity
 Generator.prototype.toEntity = function (value) {
-  // regular entity
-  if (isString(value)) {
-    switch (value[0]) {
+  if (isTerm(value)) {
+    // Remove the prefix but with taking a copy to avoid altering the input
+    if (value.termType === "BlankNode") value = {termType: value.termType, value: value.value.replace(/^e_|^g_/, '')};
```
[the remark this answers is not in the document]
simonvbrae: Actually this was not necessary anymore for the tests. It was something I added to prevent e_g_e_g_... in a query that was repeatedly parsed and then generated again. But even that won't happen with the current implementation of the parser:
`if (name.startsWith("e_")) return Parser.factory.blankNode(name);`
Blanknodes after an arbitrary amount of parsing and generation will simply be something like _:e_g_7

Are you happy with this behaviour? If so I will remove this line :)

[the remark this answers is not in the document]
RubenVerborgh: Happy with this behavior 🙂

lib/SparqlGenerator.js
```diff
@@ -31,11 +31,12 @@ Generator.prototype.toQuery = function (q) {
   if (q.distinct)
     query += 'DISTINCT ';
 
-  if (q.variables)
+  if (q.variables){
     query += mapJoin(q.variables, undefined, function (variable) {
-      return isString(variable) ? this.toEntity(variable) :
-             '(' + this.toExpression(variable.expression) + ' AS ' + variable.variable + ')';
+      return isTerm(variable) ? this.toEntity(variable) :
+             '(' + this.toExpression(variable.expression) + ' AS ' + "?" + variable.variable.value + ')';
```
rubensworks: I'm surprised the linter doesn't complain about this, but the convention in this project is to write strings with single quotes. (double quotes are used in several places throughout this PR)

rubensworks: Also, the `'?' + variable.variable.value` pattern occurs several times. I'd suggest to abstract this into a `variableToString` method.

lib/SparqlGenerator.js
```diff
@@ -358,6 +360,9 @@ Generator.prototype.indent = function(text) { return text.replace(/^/gm, this._i
 // Checks whether the object is a string
 function isString(object) { return typeof object === 'string'; }
 
+// Checks whether the object is a Term
+function isTerm(object) { return object.termType !== undefined; }
```
rubensworks:
```suggestion
function isTerm(object) { return !!object.termType; }
```

lib/Wildcard.js
```diff
@@ -0,0 +1,28 @@
...
+  get termType() {
+    return 'Wildcard';
+  }
+
+  equals(other) {
+    return other && (this.termType === other.termType);
+  }
+}
+
+Object.defineProperty(Wildcard.prototype, 'value', {
```
rubensworks: You could also do this as well for `termType`. Was there any particular reason for handling these differently?

test/SparqlGenerator-test.js
```diff
@@ -2,9 +2,12 @@ var SparqlGenerator = require('../sparql').Generator;
 var SparqlParser = require('../sparql').Parser;
 
 var fs = require('fs'),
-    expect = require('chai').expect,
+    expect = require('expect'),
     os = require('os');
 
+var toEqualParsedQuery = require("../test/matchers/toEqualParsedQuery");
+expect.extend({toEqualParsedQuery,});
```
rubensworks:
```suggestion
expect.extend({toEqualParsedQuery});
```

test/matchers/toEqualParsedQuery.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+
+// Test function which checks if object are equal, keeping into account how N3.DataFactory works
+let objectsEqual = function (received, expected){
+    if (isPrimitive(received) || received === undefined){
+        return received === expected;
+    }
+
+    if (received instanceof N3.DataFactory.internal.Term){
```
rubensworks: I suggest making this more generic, and making it work for terms from any data factory. (I foresee we may run into issues in the future because of this)
So for checking if something is a term, you could simply do `if (received.termType)`.

RubenVerborgh: I'm fine with this in tests. (We probably want to swap out N3 at a later stage, but not strictly part of this PR.)

lib/Wildcard.js
simonvbrae: Nothing in particular, will change!

RubenVerborgh:
Thanks all 🎉

RubenVerborgh:
v3.0.0-beta.0 is out!

PR #89, opened 2019-07-31
Error when GROUP BY is used in a SELECT * query

### LINKED ISSUE #84: SELECT * is not allowed with GROUP BY

```sparql
SELECT * { ?s ?p ?o } GROUP BY ?s
```

From [16.1.1 Projection](https://www.w3.org/TR/sparql11-query/#selectproject)): 

> Use of SELECT * is only permitted when the query does not have a GROUP BY clause.

Relevant test: http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_43

simonvbrae:
closes #84

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -468,13 +468,21 @@ PrefixDecl
     }
     ;
 SelectQuery
-    : SelectClause DatasetClause* WhereClause SolutionModifier -> extend($1, groupDatasets($2), $3, $4)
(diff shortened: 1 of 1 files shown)


simonvbrae:
The first commit looks weird but shouldn't be a problem! It's just a result of me updating my fork.

RubenVerborgh:
(Rebased so it's back to 1 commit.)

RubenVerborgh:
Excellent, good solution.

PR #93, opened 2019-10-02
Run SPARQL test suite on Travis builds

rubensworks:
This adds [rdf-test-suite](https://github.com/rubensworks/rdf-test-suite.js) as a dev dependency to run the SPARQL query and update test suite.
It skips all tests that execute queries, and only runs the syntax tests.

At the moment, a few query tests fail (those reported in https://github.com/comunica/comunica/issues/287), but also a few update queries fail that have not been reported before.

For the moment, I've added the `-e` flag to disable Travis build failures on test failures. Once spec compliance is restored, this flag should be removed to detect future regressions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 *.log
 lib/SparqlParser.js
+.rdf-test-suite-cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,10 @@ node_js:
 before_script:
   - npm run build
 
-cache: npm
+script:
+  - npm test
+  - npm run spec
+cache:
+ - npm
+ - directories:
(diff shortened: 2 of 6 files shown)


rubensworks:
Oops, indentation issue in yaml, fixed now.

RubenVerborgh:
Node 6 fail, but okay with dropping support.

rubensworks:
Ok, shall I also update minimum support in package.json engines, or leave as-is?

RubenVerborgh:
If you want, yes.

rubensworks:
Oh, it's already set to Node 8 there, nvm.

rubensworks:
I assume this can be merged?

RubenVerborgh:
Yes, thanks!

PR #96, opened 2019-11-23
add option to create full generator instance

lisp:
the implementation seems to be well-factored to permit an application to serialize arbitrary algebra nodes.
the interface requires just to be able to get a handle on a full generator instance.
this adds a `generator` function which yields one configured with the original options.


RubenVerborgh:
Thanks, merged as 6ca3063bf8daef1e864756d02e3a14cfcfb67762 under the name of `createGenerator`.

Issue #98, opened 2020-06-23
Update typings to RDF/JS


RubenVerborgh:
https://github.com/RubenVerborgh/SPARQL.js/issues/73#issuecomment-647374395

jacoscaz:
@RubenVerborgh I was having a look at the current typings to see whether I can make a PR for this. Looks to me like the change to be made should focus around the following definition of the  `Term` type:

```ts
export type Term = string & { __termBrand: string; };
```

Do you think it would be enough to change this to use RDF/JS' typings? Or, would it be better to also change the `Triple` interface to better express which kind of term (named node, literal, ...) can be used as the `subject` / `predicate` / `object` of a triple?

RubenVerborgh:
I think regular RDF/JS should suffice indeed, thanks!

AlexeyMz:
The new typings with RDF/JS terms are merged to DefinitelyTyped 🎉
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46159

jacoscaz:
Thank you @AlexeyMz !

PR #103, opened 2020-07-15
Fix generating triples using shorthand notation when subject/predicate is the same

AlexeyMz:
Hi @RubenVerborgh !

While updating our SPARQL.js dependency to v3.0 at [metaphacts](https://github.com/metaphacts) we discovered that some queries are generated without short notation for triple sequence with the same subject or subject-predicate.

Here is an attempt to fix it by structurally comparing RDF/JS terms. There is a helper method `equalTerms()` defined to avoid using `Term.equals()` method as it's not available when de-serializing query from JSON.

In addition we prepared correct TypeScript typings for SPARQL.js v3.0 (issue #98) which will publish soon on DefinitelyTyped 😄

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -109,11 +109,11 @@ Generator.prototype.encodeTriples = function (triples) {
   if (!triples.length)
     return '';
 
-  var parts = [], subject = '', predicate = '';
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Super, thanks a lot!

PR #104, opened 2020-07-16
Add `omitStringDatatype` option to generate simple literals for RDF literal terms with xsd:string datatype

AlexeyMz:
This PR adds ability to produce simple literals for RDF/JS Literal terms with `xsd:string` datatype.

With recent move to RDF/JS term representation in SPARQL.js v3.0 every literal created by `DataFactory` interface usually have `xsd:string` as a default datatype. This causes `Generator` to add datatype suffix for each string literal (e.g. `"foo"^^<http://www.w3.org/2001/XMLSchema#string>`) which may be a little verbose.

There is also an interesting related discussion here: https://github.com/w3c/sparql-12/issues/112

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -1,4 +1,5 @@
 var XSD_INTEGER = 'http://www.w3.org/2001/XMLSchema#integer';
+var XSD_STRING = 'http://www.w3.org/2001/XMLSchema#string';
 
 function Generator(options) {
   this._options = options = options || {};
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Shouldn't we just omit it by default? I'd suggest to have this option always on (as was the case in v2.x).

RubenVerborgh:
Currently merged as ecbb372c58a0478125a5dddde4ca14862adc42c2 (without `omitStringDatatype` flag).

AlexeyMz:
Ok, thanks a lot for a quick response! 👍

antoine-morvan:
Hi there, 

I understand this should be default, but I don't see the option to disable that. Did I miss it ? One of our virtuoso understands "PARIS"^^xsd:string but not "PARIS" :sweat_smile:

PR #106, opened 2020-07-24
Add SPARQL* support to parser and generator

### LINKED ISSUE #105: Support for SPARQL*
Dear Ruben, dear sparql.js community,
 
at [metaphacts](https://github.com/metaphacts) we are excited to see the uptake and standardization efforts of SPARQL*/RDF* within the knowledge graph community. We believe it would be a major step forward for the community to have SPARQL* also properly integrated and supported by the SPARQL.js grammar as one of the main SPARQL parsing and query manipulation libraries.

In 2018 there was already a feature request (see https://github.com/RubenVerborgh/SPARQL.js/issues/65) which was closed due to the immature state of the SPARQL*/RDF* specification and standardization effort. Since the W3C Workshop on Web Standardization for Graph Data in May 2019 in Berlin (https://www.w3.org/Data/events/data-ws-2019/cfp.html) there have been renewed efforts and many graph databases have implemented the proposed standard. Therefore we believe it is time to revisit the topic.
 
Generally, we would be happy to help implement/contribute support for SPARQL* in SPARQL.js. However, we have a few questions:

* Is this of interest to anyone and in particular the project owner?
* Is someone already working on this, i.e. avoid duplication of efforts, join forces and make plans transparent?
* What are the design- and technical prerequisites to get it properly integrated into a sparql.js release? Our current plan would be mainly to extend the parser grammar and serializer along the lines of what has been specified by Olaf Hartig in https://arxiv.org/abs/1406.3399, Section 5 “embedded triple pattern”. As far as we currently look at it, it should be a relatively small and fully compatible extension to the main sparql.js SPARQL 1.1 grammar. Recently, together with Ontotext we implemented a similar extension in RDF4J (c.f. https://github.com/eclipse/rdf4j/issues/1484 and in particular the TripleRef in the JJTree https://github.com/eclipse/rdf4j/pull/1920/commits/251c979f835af8f9f995bca85621df0d144944a2#diff-c4d16e5e970dd365f5811921f25b6a3aR1027)

For those lacking context: An excellent starting point for further reading on SPARQL*/RDF* is this (continuously updated) blog post by Olaf Hartig: https://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/ including a list of tools and example SPARQL and RDF parser implementations https://github.com/RDFstar/RDFstarTools .

 
Cheers,
Johannes on behalf of the [metaphacts](https://github.com/metaphacts) Team

AlexeyMz:
This is a draft implementation of SPARQL* support in parser and generator from our team at metaphacts.
(It fixes #105)

I'm not very familiar with parser rules structure but I tried my best at making as little changes as possible to support triple references (`<<s p o>>`).
Also the generator currently explicitly checks if a term is a quad which won't be needed in the future (as `Quad` will implement `Term` interface).

### DIFF
--- a/bin/sparql-to-json
+++ b/bin/sparql-to-json
@@ -16,5 +16,20 @@ source.setEncoding('utf8');
 source.on('data', function (data) { query += data; });
 source.on('end', function () {
   var parseTree = new SparqlParser().parse(query);
-  process.stdout.write(JSON.stringify(parseTree, null, '  ') + '\n');
+  function termReplacer(key, value) {
+    if (isQuad(value)) {
+      // Workaround for N3 Quad.toJSON() missing `termType` property
+      return {termType: 'Quad', ...value};
+    }
+    return value;
+  };
+  process.stdout.write(JSON.stringify(parseTree, termReplacer, '  ') + '\n');
 });
+
+function isQuad(value) {
+  return value && typeof value === 'object'
+    && !!value.subject
+    && !!value.predicate
+    && !!value.object
+    && !!value.graph;
+}
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -285,6 +285,14 @@ Generator.prototype.toEntity = function (value) {
         value += '^^' + this.encodeIRI(datatype.value);
       }
       return value;
+    case 'Quad':
+      return (
+        '<< ' +
+        this.toEntity(value.subject) + ' ' +
+        this.toEntity(value.predicate) + ' ' +
+        this.toEntity(value.object) +
+        ' >>'
+      );
     // IRI
     default:
       return this.encodeIRI(value.value);
@@ -374,7 +382,9 @@ function variableToString(variable){
 function isString(object) { return typeof object === 'string'; }
 
 // Checks whether the object is a Term
-function isTerm(object) { return !!object.termType; }
+function isTerm(object) {
+  return !!object.termType;
+}
 
 // Checks whether term1 and term2 are equivalent without `.equals()` prototype method
 function equalTerms(term1, term2) {
@@ -386,6 +396,11 @@ function equalTerms(term1, term2) {
       return term1.value === term2.value
         && term1.language === term2.language
         && equalTerms(term1.datatype, term2.datatype);
+    case 'Quad':
+      return equalTerms(term1.subject, term2.subject)
+        && equalTerms(term1.predicate, term2.predicate)
+        && equalTerms(term1.object, term2.object)
+        && equalTerms(term1.graph, term2.graph);
     default:
       return term1.value === term2.value;
   }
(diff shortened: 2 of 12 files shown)


rubensworks:
> N3.js can also work with triples that have quads as their subject. For example <<<a> <b> <c> <g>>> <b> <e>. I suggest adapting SPARQL to this as well.

We may want to add a config option for this. Default RDF* does not allow quads here, but when enabled via a config option we could allow this, as this may be valuable for experimentation.

Also, we may want to follow the SPARQL syntax for representing quads here, i.e. `<<GRAPH <g> { <a> <b> <c> }>> <b> <e>`

RubenVerborgh:
Also, it would need to be behind a flag for me, such that we can be 100% compatible with SPARQL 1.1 (i.e., reject queries tthat are not valid SPARQL 1.1, which includes SPARQL*).

AlexeyMz:
@simonvbrae @rubensworks I agree, this would be a valuable feature to experiment with. For now I think it would be great to have a basic syntax support without nested triples and quads; it should be easy to extend it in the near future under a flag.

@RubenVerborgh I updated the PR to use N3 v1.6.0 with RDF* support and added `allowRdfStar` flag to the parser to throw an error on `<< >>` syntax if it's not explicitly set.

simonvbrae:
> @simonvbrae @rubensworks I agree, this would be a valuable feature to experiment with. For now I think it would be great to have a basic syntax support without nested triples and quads; it should be easy to extend it in the near future under a flag.
> 
> @RubenVerborgh I updated the PR to use N3 v1.6.0 with RDF* support and added `allowRdfStar` flag to the parser to throw an error on `<< >>` syntax if it's not explicitly set.

I have implemented nested triples and tests. I could add this to your code and then, if @RubenVerborgh approves, this PR should be good to go.

AlexeyMz:
@simonvbrae It would be great if you update this PR with nested triples support 😄

bin/sparql-to-json
```diff
@@ -16,5 +16,20 @@ source.setEncoding('utf8');
 source.on('data', function (data) { query += data; });
 source.on('end', function () {
   var parseTree = new SparqlParser().parse(query);
-  process.stdout.write(JSON.stringify(parseTree, null, '  ') + '\n');
+  function termReplacer(key, value) {
+    if (isQuad(value)) {
+      // Workaround for N3 Quad.toJSON() missing `termType` property
```
RubenVerborgh, line 21: @simonvbrae Do we need to fix this in N3?

RubenVerborgh:
Ah wait, we probably want to continue this in #111 instead.

bin/sparql-to-json
simonvbrae: Hmm I thought so too initially. I checked to see if I could find the problem in N3 but I didn't see a problem there... @AlexeyMz Why do we need this workaround?

PR #107, opened 2020-07-28
Check if select_vars appear in group clause

### LINKED ISSUE #83: Out-of-scope variables in SELECT not throwing error

```sparql
SELECT ?o WHERE { ?s ?p ?o } GROUP BY ?s
```

Relevant test: http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_44

Test message:

> required syntax error : out of scope variable in SELECT from group

FlorianFV:
Closes #83 
Fixes following tests of test suite:

[x] syn-bad-02.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_44)
[x] COUNT 9 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg09) (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_44) 
[x] COUNT 12 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg12)
[x] Group-6 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/grouping/manifest#group06)
[x] Group-7 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/grouping/manifest#group07)

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -266,6 +266,11 @@
     }
     return merged;
   }
+
+  // Return the id of an expression
+  function getIdOfExpression(expression) {
+    return expression.variable ? expression.variable.id : expression.id || expression.expression.id;
+  }
 %}
 
 %lex
@@ -472,7 +477,21 @@ PrefixDecl
     ;
 SelectQuery
     : SelectClauseWildcard DatasetClause* WhereClause SolutionModifierNoGroup -> extend($1, groupDatasets($2), $3, $4)
-    | SelectClauseVars     DatasetClause* WhereClause SolutionModifier        -> extend($1, groupDatasets($2), $3, $4)
+    | SelectClauseVars     DatasetClause* WhereClause SolutionModifier 
+    {
+      // If there is a group clause, check if all variables that are selected are in the grouped clause
+      if ($4.group) {
+        for (let selectVar of $1.variables) {
+          if (!selectVar.expression) {
+            const selectVarId = getIdOfExpression(selectVar);
+            if ($4.group.filter(groupVar => getIdOfExpression(groupVar) === selectVarId).length === 0) {
+              throw Error("select variable out of scope");
+            }
+          }
+        }
+      } 
+      return extend($1, groupDatasets($2), $3, $4);
+    }
     ;
 SelectClauseWildcard
     : SelectClauseBase '*' -> extend($1, {variables: [new Wildcard()]})


lib/sparql.jison
```diff
@@ -266,6 +266,11 @@
     }
     return merged;
   }
+
+  // Return the id of an expression
+  function getIdOfExpression(expression) {
+    return expression.variable ? expression.variable.id : expression.id || expression.expression.id;
```
RubenVerborgh, line 272: Just wondering if that last case is always sufficient; or do we want:

```suggestion
    return expression.variable ? expression.variable.id : expression.id || getIdOfExpression(expression.id);
```

Also, the nesting of `? : ||` might be confusing; parentheses could help.

lib/sparql.jison
```diff
@@ -266,6 +266,11 @@
     }
     return merged;
   }
+
+  // Return the id of an expression
+  function getIdOfExpression(expression) {
```
RubenVerborgh, line 271: `getExpressionId`?

lib/sparql.jison
```diff
@@ -472,7 +477,21 @@ PrefixDecl
     ;
 SelectQuery
     : SelectClauseWildcard DatasetClause* WhereClause SolutionModifierNoGroup -> extend($1, groupDatasets($2), $3, $4)
-    | SelectClauseVars     DatasetClause* WhereClause SolutionModifier        -> extend($1, groupDatasets($2), $3, $4)
+    | SelectClauseVars     DatasetClause* WhereClause SolutionModifier 
+    {
+      // If there is a group clause, check if all variables that are selected are in the grouped clause
+      if ($4.group) {
+        for (let selectVar of $1.variables) {
```
RubenVerborgh, line 484:
```suggestion
        for (const selectVar of $1.variables) {
```

lib/sparql.jison
```diff
@@ -472,7 +477,21 @@ PrefixDecl
...
+          if (!selectVar.expression) {
+            const selectVarId = getIdOfExpression(selectVar);
+            if ($4.group.filter(groupVar => getIdOfExpression(groupVar) === selectVarId).length === 0) {
```
RubenVerborgh, line 487: Instead of filtering and checking for length, we can just use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find (which would stop after the first match, and not create an array)

lib/sparql.jison
FlorianFV: We actually want the id of the outer expression. The first case is that it is indeed an expression with an 'AS' (explicit like: ?O1 + ?O2) AS ?O12 or implicit like: count(?O) where a free variable name is added to the expression). The second case is for a variable, which directly has an id field, but can also be packed as an expression, so then it is in the expression field and then we are sure there is just a variable with an id field.

RubenVerborgh: Sounds good, perhaps add a quick comment with this?

PR #108, opened 2020-07-28
Check for projection of ungrouped variables

### LINKED ISSUE #83: Out-of-scope variables in SELECT not throwing error

### LINKED ISSUE #91: Group by with aggregate should be invalid

```sparql
PREFIX : <http://www.example.org/>

SELECT ((?O1 + ?O2) AS ?O12) (COUNT(?O1) AS ?C)
WHERE { ?S :p ?O1; :q ?O2 } GROUP BY (?O1 + ?O2)
ORDER BY ?O12
```

Relevant test: http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg08

Test message:

> grouping by expression, done wrong

(To be honest, I don't immediately see the reason why this should fail. Something to look into...)

**EDIT**: This may actually be related to #83, as project expressions can only use grouped variables, while `?O1` and `?O2` are being used.

FlorianFV:
Closes #83 
Closes #91 
Fixes following tests of test suite:

[x] syn-bad-02.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_44)
[x] COUNT 8 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg08)
[x] COUNT 9 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg09) (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_44) 
[x] COUNT 10 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg10)
[x] COUNT 11 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg11)
[x] COUNT 12 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/aggregates/manifest#agg12)
[x] Group-6 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/grouping/manifest#group06)
[x] Group-7 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/grouping/manifest#group07)

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -281,7 +281,7 @@
       return [expression];
     } else if (expression.type === "operation") {
       const aggregates = [];
-      for (let arg of expression.args) {
+      for (const arg of expression.args) {
         aggregates.push(...getAggregatesOfExpression(arg));
       }
       return aggregates;
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -541,7 +541,7 @@ SelectQuery
           }
         }
       }
-      return extend($1, groupDatasets($2), $3, $4)
+      $$ = extend($1, groupDatasets($2), $3, $4)
     }
     ;
 SelectClauseWildcard
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -268,7 +268,7 @@
   }
 
   // Return the id of an expression
-  function getIdOfExpression(expression) {
+  function getExpressionId(expression) {
     return expression.variable ? expression.variable.id : expression.id || expression.expression.id;
   }
 
@@ -280,9 +280,9 @@
     if (expression.type === 'aggregate') {
       return [expression];
     } else if (expression.type === "operation") {
-      let aggregates = [];
+      const aggregates = [];
       for (let arg of expression.args) {
-        aggregates = aggregates.concat(getAggregatesOfExpression(arg));
+        aggregates.push(...getAggregatesOfExpression(arg));
       }
       return aggregates;
     }
@@ -291,21 +291,21 @@
 
   // Get all variables used in an operation
   function getVariablesFromOperation(operation) {
-    let variables = [];
-    for (let arg of operation.args) {
-      if (arg.id) {
-        variables.push(arg);
+    const variables = new Set();
+    for (const arg of operation.args) {
+      if (arg.termType === "Variable") {
+        variables.add(arg);
       } else if (arg.type === "operation") {
-        variables = variables.concat(getVariablesFromOperation(arg));
+        getVariablesFromOperation(arg).forEach(v => variables.add(v));
       }
     }
     return variables;
   }
 
-  // Help function to flatten arrays
+  // Helper function to flatten arrays
   function flatten(input, depth = 1, stack = []) {
-    for (let item of input) {
-        if (item instanceof Array && depth > 0) {
+    for (const item of input) {
+        if (depth > 0 && item instanceof Array) {
           flatten(item, depth - 1, stack);
         } else {
           stack.push(item);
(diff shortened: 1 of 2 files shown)


lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
     }
     return merged;
   }
+
```
RubenVerborgh, line 269: Also see review https://github.com/RubenVerborgh/SPARQL.js/pull/107#pullrequestreview-457013041

lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
...
+  function getAggregatesOfExpression(expression) {
+    if (!expression) {
+      return [];
+    }
+    if (expression.type === 'aggregate') {
+      return [expression];
+    } else if (expression.type === "operation") {
+      let aggregates = [];
+      for (let arg of expression.args) {
+        aggregates = aggregates.concat(getAggregatesOfExpression(arg));
```
RubenVerborgh, line 285: or

```suggestion
        aggregates = [...aggregates, ...getAggregatesOfExpression(arg))];
```

or even

```suggestion
        aggregates.push(...getAggregatesOfExpression(arg));
```

lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
...
+  // Get all "aggregate"'s from an expression
+  function getAggregatesOfExpression(expression) {
+    if (!expression) {
+      return [];
+    }
+    if (expression.type === 'aggregate') {
+      return [expression];
+    } else if (expression.type === "operation") {
+      let aggregates = [];
+      for (let arg of expression.args) {
```
RubenVerborgh, line 284:
```suggestion
      for (const arg of expression.args) {
```

lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
...
+      if (arg.id) {
+        variables.push(arg);
+      } else if (arg.type === "operation") {
+        variables = variables.concat(getVariablesFromOperation(arg));
+      }
+    }
+    return variables;
+  }
+
+  // Help function to flatten arrays
```
RubenVerborgh, line 305:
```suggestion
  // Helper function to flatten arrays
```

lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
...
+  function flatten(input, depth = 1, stack = []) {
+    for (let item of input) {
```
RubenVerborgh, line 307:
```suggestion
    for (const item of input) {
```

lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
...
+        if (item instanceof Array && depth > 0) {
```
RubenVerborgh, line 308: (minor) put the depth check first, as the other one is more expensive

lib/sparql.jison
```diff
@@ -472,7 +520,29 @@ PrefixDecl
...
+    | SelectClauseVars     DatasetClause* WhereClause SolutionModifier
+    {
+      // Check for projection of ungrouped variable
+      const counts = flatten($1.variables.map(vars => getAggregatesOfExpression(vars.expression)))
+        .filter(agg => agg.aggregation === "count");
+      if (counts.length > 0 || $4.group) {
+        for (const selectVar of $1.variables) {
+          if (selectVar.id) {
+            if (!$4.group || !$4.group.map(groupVar => getIdOfExpression(groupVar)).includes(selectVar.id)) {
+              throw Error("Projection of ungrouped variable");
```
RubenVerborgh, line 532: Let's  detail which one

lib/sparql.jison
```diff
@@ -472,7 +520,29 @@ PrefixDecl
...
+            }
+          } else if (getAggregatesOfExpression(selectVar.expression).length === 0) {
+            const usedVars = getVariablesFromOperation(selectVar.expression);
+            for (const usedVar of usedVars) {
+              if (!$4.group.map(groupVar => getIdOfExpression(groupVar)).includes(usedVar.id)) {
+                throw Error("Use of ungrouped variable in projection of operation");
```
RubenVerborgh, line 538: Let's  detail which one

lib/sparql.jison
```diff
@@ -266,6 +266,54 @@
     }
     return merged;
   }
+
+  // Return the id of an expression
+  function getIdOfExpression(expression) {
+    return expression.variable ? expression.variable.id : expression.id || expression.expression.id;
```
RubenVerborgh, line 272: `.id` is implementation-specific; @rubensworks, what would you recommend? N3.js exposes `termToId` if needed (but I don't think it is)

rubensworks: I'm not that familiar with this codebase, but it looks like only variables (or variables in expressions) are passed here (but not sure). In that case, I guess we could just use `.value` instead of `.id`.

If non-variable terms are also applicable here, then [`termToString`](https://github.com/rubensworks/rdf-string.js#term-to-string) might be the safest solution here.

FlorianFV: The variables the parser gives me are in this shape: `Variable { id: '?x' }` . There is no value field availble (it's undefined if I print it), which makes the .value and termToString option not possible).
Expressions are of the following form (as I get them from the parser):  `{ expression: Variable { id: '?x' } }` . When using `termToId` of N3.js, it shows the following error: `Unexpected termType: undefined` . I'm not entirely sure what to do now.

rubensworks: `.value` on those terms should be possible though. It's not because they are not printed when calling `console.log(term)`, that they are not there. Not all fields in JS objects are necessarily printed.

An expression is not an RDF/JS term AFAIK, so a crash when calling term-operations on it is expected. So if you want to determine some unique id for those expressions, you'll have to look for a workaround, possibly by checking what this expression is defined over.

PR #110, opened 2020-07-29
Fix parsing subtraction expressions (fixes regression after switch to RDF/JS literals)

### LINKED ISSUE #101: Arithmetic expression fails to parse
This successfully parses:

```
SELECT * WHERE {
BIND (?a+1 AS ?b)
}
```

where this doesnt:

```
SELECT * WHERE {
BIND (?a-1 AS ?b)
}
```

That last part throws the following error:

```
     TypeError: $$[($0 - 1)].replace is not a function
      at Object.anonymous (/home/lrd900/triply/core/node_modules/sparqljs/lib/SparqlParser.js:429:45)
      at Parser.parse (/home/lrd900/triply/core/node_modules/sparqljs/lib/SparqlParser.js:660:36)
      at Parser.parser.parse [as parse] (/home/lrd900/triply/core/node_modules/sparqljs/sparql.js:25:37)
```

AlexeyMz:
Fixes #101

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -131,7 +131,7 @@
     return group;
   }
 
-  // Converts the number to a string
+  // Converts the string to a number
   function toInt(string) {
     return parseInt(string, 10);
   }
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Thanks so much!

PR #111, opened 2020-07-29
Add SPARQL* support

simonvbrae:
This pull request adds SPARQL* support to the parser and generator. It also adds an argument for specifying whether SPARQL* should be allowed or not. The implementation was based on @AlexeyMz's contribution, discussion and initial contribution can be found at PR #106.

Details:
- [x] Support for SPARQL*
- [x] Support for nested triples
- [x] Tests
- [x] Commandline flag and Parser argument `--strict` to specify SPARQL* support
- [ ] Support for quads in the place of triples
- [ ] Update README to reflect the `--strict` argument

### DIFF
--- a/bin/sparql-to-json
+++ b/bin/sparql-to-json
@@ -22,6 +22,6 @@ var query = '', source = args[0] ? fs.createReadStream(args[0]) : process.stdin;
 source.setEncoding('utf8');
 source.on('data', function (data) { query += data; });
 source.on('end', function () {
-  var parseTree = new SparqlParser({ strictMode }).parse(query);
+  var parseTree = new SparqlParser({ sparqlStar: !strictMode }).parse(query);
   process.stdout.write(JSON.stringify(parseTree, null, '  ') + '\n');
 });
--- a/bin/sparql-to-json
+++ b/bin/sparql-to-json
@@ -2,9 +2,16 @@
 
 // Parse arguments
 var args = process.argv.slice(2);
+let inStrictMode = false;
 if (args.length > 1) {
-  console.error('usage: sparql-to-json query.sparql');
-  return process.exit(1);
+  if (args[0] !== '--strict') {
+    console.error('usage: sparql-to-json query.sparql');
+    return process.exit(1);
+  } else {
+    inStrictMode = true;
+    // Shift arguments
+    args.shift();
+  }
 }
 
 var fs = require('fs'),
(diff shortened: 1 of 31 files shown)


bin/sparql-to-json
```diff
@@ -2,9 +2,16 @@
 
 // Parse arguments
 var args = process.argv.slice(2);
+let inStrictMode = false;
```
rubensworks, line 5: I think @RubenVerborgh implied that SPARQL* should only be possible when it has been explicitly enabled. So creating a parser _without_ this option defined should reject SPARQL* queries.

lib/sparql.jison
```diff
@@ -266,6 +266,13 @@
     }
     return merged;
   }
+
+  function ifNotStrictMode(value) {
+    if (Parser.inStrictMode) {
+      throw new Error('RDF* is not allowed in strict mode');
```
rubensworks, line 272: In this context, it should be named SPARQL* instead of RDF*.

bin/sparql-to-json
RubenVerborgh: I'm fine the command-line being lenient by default though.

RubenVerborgh:
```suggestion
let strictMode = false;
```

lib/SparqlGenerator.js
```diff
@@ -374,7 +382,9 @@ function variableToString(variable){
 function isString(object) { return typeof object === 'string'; }
 
 // Checks whether the object is a Term
-function isTerm(object) { return !!object.termType; }
+function isTerm(object) {
+  return !!object.termType;
```
RubenVerborgh, line 386:
```suggestion
  return typeof object.termType === 'string';
```

bin/sparql-to-json
simonvbrae: It's only the commandline version that parses SPARQL* by default, the exported parser still rejects RDF* by default. This was Ruben's suggestion. But it was discussed privately in Mattermost.

rubensworks: @simonvbrae You'll probably want to refer to _Ruben_ by `@`, especially when there are multiple of them in the same discussion 😛

RubenVerborgh: There can be only one.

simonvbrae: This is going to get very confusing :p

RubenVerborgh:
Good to go after @simonvbrae adds quick docs to the README.

RubenVerborgh:
Thanks a lot, @AlexeyMz and @simonvbrae!!

PR #113, opened 2020-07-30
Improve lexer performance on invalid queries by optimizing Regex definitions

### LINKED ISSUE #41: Parsing simple erroneous queries can take exponential time
@JonasKress noticed that if you simply type `ddddddddddddddddddddddddddddddddd` (or any other letter) on query.wikidata.org, the tab hangs. It seems that SPARQL.js takes exponential time (twice as long for each letter) to parse such a “query” (before reporting a parse error):

```bash
$ for i in {20..30}; do start=$SECONDS; printf '%*s' $i | tr ' ' 'x' | sparqljs &>/dev/null; end=$SECONDS; printf '%2d: %d s\n' $i $((end-start)); done
20: 0 s
21: 0 s
22: 1 s
23: 1 s
24: 2 s
25: 3 s
26: 7 s
27: 14 s
28: 28 s
29: 56 s
30: 112 s
```

Do you think it’s possible to improve this?

AlexeyMz:
This PR turns Regex capturing groups into non-capturing (`(?:)`) and combines alternatives (`|`) into a single character set (`[a-z]`) if possible.

Fixes #41 (at least looks like it).

EDIT1: I just noticed there is also an error in the original `PN_CHARS_BASE` definition (a missing `|`) which is fixed by this.
EDIT2: As it turns out, this wasn't an error but an explicit support for UTF-16 surrogate pairs, so this part was reverted back and I added a test.


RubenVerborgh:
That is amazing, thank you!

PR #114, opened 2020-07-30
Fix parsing boolean values with different character case

### LINKED ISSUE #92: Error on parsing boolean
The following code easily reproduce the issue.

```
const SparqlJS = require("sparqljs");

const SparqlParser = SparqlJS.Parser;
const SparqlGenerator = SparqlJS.Generator;

let parser = new SparqlParser();
let generator = new SparqlGenerator();


let query = "SELECT ?s ?p ?o { ?s ?p True }";

let queryObj = parser.parse(query);

let newQuery = generator.stringify(queryObj);

console.log(newQuery);

```

The output:
```
SELECT ?s ?p ?o WHERE { ?s ?p "True"^^<http://www.w3.org/2001/XMLSchema#boolean>. }
```

According to the specification of xml schema (https://www.w3.org/TR/xmlschema-2/#boolean), boolean values should be `{true|false}`. Therefore, The literal `"True"^^<http://www.w3.org/2001/XMLSchema#boolean>.` is wrong, which value is set by mistake of `True` in the sparql. For instance, this sample query evaluates differently the original one on Jena.

AlexeyMz:
Fixes #92

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -929,7 +929,7 @@ Literal
     | DOUBLE  -> createTypedLiteral(lowercase($1), XSD_DOUBLE)
     | NumericLiteralPositive
     | NumericLiteralNegative
-    | BOOLEAN -> createTypedLiteral($1, XSD_BOOLEAN)
+    | BOOLEAN -> createTypedLiteral($1.toLowerCase(), XSD_BOOLEAN)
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Excellent, thank you!

PR #115, opened 2020-07-31
SPARQL* followup: allow variable BIND/projection, constant triple patterns in VALUES()

AlexeyMz:
According to the paper it should be allowed to use variables inside triple patterns when encountered in `BIND` or projection clauses and constant triple patterns inside `VALUES()`: https://arxiv.org/pdf/1406.3399.pdf

This is also how rdf4j implements it: https://github.com/eclipse/rdf4j/pull/1920/commits/251c979f835af8f9f995bca85621df0d144944a2#diff-c4d16e5e970dd365f5811921f25b6a3aR1027

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -623,7 +623,8 @@ SubSelect
     ;
 SelectClauseItem
     : VAR -> toVar($1)
-    | '(' (Expression | VarTriple) 'AS' VAR ')' -> expression($2, { variable: toVar($4) })
+    | '(' Expression 'AS' VAR ')' -> expression($2, { variable: toVar($4) })
+    | '(' VarTriple 'AS' VAR ')' -> ensureSparqlStar(expression($2, { variable: toVar($4) }))
     ;
 ConstructQuery
     : 'CONSTRUCT' ConstructTemplate DatasetClause* WhereClause SolutionModifier -> extend({ queryType: 'CONSTRUCT', template: $2 }, groupDatasets($3), $4, $5)
@@ -705,7 +706,7 @@ InlineData
 DataBlockValue
     : iri
     | Literal
-    | ConstTriple
+    | ConstTriple -> ensureSparqlStar($1)
     | 'UNDEF' -> undefined
     ;
 DataBlockValueList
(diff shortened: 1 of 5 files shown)


lib/sparql.jison
```diff
@@ -567,7 +567,7 @@ SubSelect
     ;
 SelectClauseItem
     : VAR -> toVar($1)
-    | '(' (Expression | ConstTriple) 'AS' VAR ')' -> expression($2, { variable: toVar($4) })
+    | '(' (Expression | VarTriple) 'AS' VAR ')' -> expression($2, { variable: toVar($4) })
```
RubenVerborgh, line 570: We'll need an `ensureSparqlStar` around some of that (fc204e9f57564b56d7054fc79b005319ecc58c52); perhaps need to split the rule (but avoid if possible, that's more expensive).

lib/sparql.jison
```diff
@@ -649,6 +649,7 @@ InlineData
 DataBlockValue
     : iri
     | Literal
+    | ConstTriple
```
RubenVerborgh, line 652: Idem, `ensureSparqlStar`.

lib/sparql.jison
```diff
@@ -730,7 +731,7 @@ GraphPatternNotTriples
     | 'GRAPH' (VAR | iri) GroupGraphPattern -> extend($3, { type: 'graph', name: toVar($2) })
     | 'SERVICE' 'SILENT'? (VAR | iri) GroupGraphPattern -> extend($4, { type: 'service', name: toVar($3), silent: !!$2 })
     | 'FILTER' Constraint -> { type: 'filter', expression: $2 }
-    | 'BIND' '(' (Expression | ConstTriple) 'AS' VAR ')' -> { type: 'bind', variable: toVar($5), expression: $3 }
+    | 'BIND' '(' (Expression | VarTriple) 'AS' VAR ')' -> { type: 'bind', variable: toVar($5), expression: $3 }
```
RubenVerborgh, line 734: Idem, `ensureSparqlStar`.

lib/sparql.jison
AlexeyMz: Should we need to call `ensureSparqlStar` here in addition to `VarTriple` / `ConstTriple` rule? I think it will fail when matching these rules first.

RubenVerborgh: Yeah; this is why we might want to split this into two rules.

AlexeyMz:
@RubenVerborgh I updated the PR (added `ensureSparqlStar` calls where needed).

RubenVerborgh:
Thanks so much!

PR #122, opened 2020-11-03
Replaces n3.DataFactory with rdf-data-factory (closes #121)

### LINKED ISSUE #121: Use smaller data factory dependency
@jacoscaz is working on a custom Comunica config for [Quadstore](https://github.com/beautifulinteractions/node-quadstore), where a priority is to have small bundle sizes.

After some improvements in Comunica, this shows the largest current dependencies:
Screen-Shot-2020-11-02-at-15 55 46

N3 seems to be quite large here, but actually isn't needed.
N3 is only used here as a dependency of SPARQL.js, since it's parsers and serializers were excluded from the Comunica config.
SPARQL.js only seems to use N3 for its data factory implementation.

This issue is to discuss the possibility to migrate to another (smaller) data factory dependency.
For instance, [rdf-data-factory](https://github.com/rubensworks/rdf-data-factory.js) could be used, as this dependency is already in use by all other Comunica dependencies. (It's the smallest data factory AFAIK)

jacoscaz:
As per #121 , this PR replaces `n3.DataFactory` to a stand-alone implementation of the RDF/JS `DataFactory` interface to reduce the overall bundle size of Comunica for those configurations that do not make use of `n3`'s parsers and/or serializers.

A few notes:

- The `rdf-data-factory` package does not export a shared `Term` abstract class the same way `n3` does. This is what motivated the change to `Wildcard` but most importantly to `toEqualParsedQuery.js`, where I've had to add the `isTerm()` function.
- The latter could perhaps be improved, performance wise, by doing something like `return value && (value instanceof Quad || value.termType)` but that felt a little hackish and obscure, so I opted for multiple `instanceof`s.

### DIFF
--- a/test/matchers/toEqualParsedQuery.js
+++ b/test/matchers/toEqualParsedQuery.js
@@ -1,5 +1,6 @@
 const diff = require('jest-diff');
-const { Term } = require('n3');
+const {Wildcard} = require('../../lib/Wildcard');
+const {BlankNode, Variable, NamedNode, DefaultGraph, Quad} = require('rdf-data-factory');
 
 function toEqualParsedQuery(received, expected) {
     const options = {
@@ -43,9 +44,9 @@ let objectsEqual = function (received, expected){
         return received === expected;
     }
 
-    if (received instanceof Term){
+    if (isTerm(received)) {
         return received.equals(expected);
-    } else if (expected instanceof Term){
+    } else if (isTerm(expected)) {
         return expected.equals(received);
     } else {
         if (Array.isArray(received)){
@@ -65,6 +66,15 @@ let objectsEqual = function (received, expected){
         return true;
     }
 };
+
+function isTerm(value) {
+    return value instanceof DefaultGraph
+        || value instanceof NamedNode
+        || value instanceof BlankNode
+        || value instanceof Variable
+        || value instanceof Wildcard
+        || value instanceof Quad;
+}
 
 function isPrimitive(value){
     return typeof value === "string" || typeof value === "number" || typeof value === "boolean";
--- a/lib/Wildcard.js
+++ b/lib/Wildcard.js
@@ -1,9 +1,7 @@
-const { Term } = require('n3');
 
 // Wildcard constructor
-class Wildcard extends Term {
+class Wildcard {
   constructor() {
-    super('');
     return WILDCARD || this;
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,43 +74,14 @@
         "@types/yargs": "^13.0.0"
       }
     },
-    "@rdfjs/data-model": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/@rdfjs/data-model/-/data-model-1.1.2.tgz",
-      "integrity": "sha512-pk/G/JLYGaXesoBLvEmoC/ic0H3B79fTyS0Ujjh5YQB2DZW+mn05ZowFFv88rjB9jf7c1XE5XSmf8jzn6U0HHA==",
+    "@types/http-link-header": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@types/http-link-header/-/http-link-header-1.0.2.tgz",
+      "integrity": "sha512-rWvCGMtwx+01LFVpLbSYagloSMgqDwfzLSx9JcwUEkJWo4oDBKihp6X92Ff5tIS4VE5ojV4wH6iMnAnr/TZhhg==",
       "dev": true,
       "requires": {
-        "@types/rdf-js": "^2.0.1"
-      },
-      "dependencies": {
-        "@types/rdf-js": {
-          "version": "2.0.12",
-          "resolved": "https://registry.npmjs.org/@types/rdf-js/-/rdf-js-2.0.12.tgz",
-          "integrity": "sha512-NBzHFHp2vHOJkPlSqzsOrkEsD9grKn+PdFjZieIw59pc0FlRG6WEQAjQZvHzFxJlYzC6ZDCFyTA1wBvUnnzUQw==",
-          "dev": true,
-          "requires": {
-            "@types/node": "*"
-          }
-        }
+        "@types/node": "*"
       }
-    },
-    "@types/color-name": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/@types/color-name/-/color-name-1.1.1.tgz",
-      "integrity": "sha512-rr+OQyAjxze7GgWrSaJwydHStIhHq2lvY3BOC2Mj7KnzI7XK0Uw1TOOdI9lDoajEbSWLiYgoo4f1R51erQfhPQ==",
-      "dev": true
-    },
-    "@types/http-link-header": {
-      "version": "1.0.1",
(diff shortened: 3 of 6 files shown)


test/matchers/toEqualParsedQuery.js
```diff
@@ -66,6 +67,15 @@ let objectsEqual = function (received, expected){
     }
 };
 
+function isTerm(value) {
+    return value instanceof DefaultGraph
```
rubensworks, line 71: Probably doesn't matter too much since this is just used for unit testing, but something like `'termType' in value` might be simpler.

jacoscaz: I tried that but using something like `return 'termType' in value` leads to a significant number of errors similar to `other.datatype.equals is not a function`. Adding more checks started to look really ugly, so I opted to say with multiple `instanceof`s.

RubenVerborgh (approved):
I'm happy; just awaiting @rubensworks' final judgement before merging.
I think we do want to release a beta though and test it with a project like Comunica to see if everything works as expected.

rubensworks (approved):
Looks good to me as well!
My earlier comment was just a thought, but it appears to not be usable here, so you can ignore it :-)

jacoscaz:
+1 on the beta release. I should be able to test it within quadstore, too.

RubenVerborgh:
v3.2.0-beta.0 released

jacoscaz:
Would it be worth releasing a beta of `SPARQLAlgebra.js`, too?

RubenVerborgh:
@joachimvh?

joachimvh:
Just did a quick run and all tests still pass with this version in sparqlalgebrajs. The only thing that is required is to release a new version with this dependency so it can be tested in Comunica?

rubensworks:
I just tested in Comunica, and there seem to be a bunch of breakages there.
Failure of 16 unit tests, 221 spec tests, and all integration tests.

rubensworks:
A quick look seems to indicate problems in sparqlalgebrajs, such as:
```
  Error: Expected not throw an error when parsing.
  Input: SELECT *
WHERE {
  {SELECT (1 AS ?X ) {}
  }
  {SELECT (1 AS ?X ) {}
  }
}

  Error: Error: Unknown expression: "\"1\"^^http://www.w3.org/2001/XMLSchema#integer"
```

joachimvh:
That is strange that it suddenly starts throwing that. Does mean that it might be more complicated to update since apparently none of the tests there cover a case like this.

rubensworks:
This should be backwards-compatible though. So it should not need changes in sparqlalgebrajs.
I suspect something may be going wrong with the `Wildcard` class.

jacoscaz:
Just tested `3.2.0-beta.0` with `quadstore`, everything seems to work all right but our tests focus on the persistence side of things and we don't have anything like that query. So, I briefly added a test with the same query 

```sparql
      SELECT *
        WHERE {
          { SELECT (1 AS ?X ) {} }
          { SELECT (1 AS ?X ) {} }
        }
```

and used `npm link` to point `quadstore` to a local version of `sparqlalgebrajs` whose dependency on `sparqljs` I updated to `3.2.0-beta.0`. Weirdly, the test seems to be passing without issues. I've double-checked with `head -n 10 nodes_modules/sparqljs/package.json` to make sure I was hitting the correct version and it does look like the case.

rubensworks:
@jacoscaz Is that test including comunica and sparqlalgebrajs? Or is it directly using sparqljs?
Perhaps the problem only occurs in combination with sparqlalgebrajs.

RubenVerborgh:
It might be a problem of running multiple SPARQL.js versions at the same time? (Still a problem though.)

jacoscaz:
@rubensworks `quadstore -> sparqlalgebrajs -> sparqljs` . I had expected the test to fail! I'll try to reproduce again later and double-check again.

rubensworks:
Weird, that should be the same as on my end.
Could someone else try linking in Comunica, just to make sure it's not just something broken on my machine?

joachimvh:
`yarn run spec` in `actor-init-sparql` passed all 284 tests for me. 12 unit tests failed to run with `yarn run test` in the root. But from what I can see that is because typescript is complaining about certain non-strict assignment in sparqlalgebrajs (e.g. assigning undefined to a boolean) since comunica is set to strict but sparqlalgebrajs isn't. I did have to screw around a bit to get everything working so please re-check 😄 .

jacoscaz:
I was able to reproduce @joachimvh ’s findings, although I, too, had to fiddle around quite a lot. I ended up creating those symlinks by hand.

jacoscaz:
I tried something slightly different this morning:

1. cleared everything from yesterday's tinkering
2. cloned `sparqlalgebrajs`, modified its `package.json` to point to `sparljs@3.2.0-beta.0`, run `npm install`
3. verified all tests passing on the modified `sparqlalgebrajs` by running `npm test`
4. run `yarn link` on the modified `sparqlagebrajs`
5. run `yarn link` on `sparqlalgebrajs/node_modules/sparqljs`
5. cloned `comunica`, run `yarn`, run `yarn link sparqlalgebrajs`, run `yarn link sparqljs`
6. verified `comunica/node_modules/sparqlalgebrajs` to be a softlink to `sparqlalgebrajs`
7. verified `comunica/node_modules/sparqljs` to be a softlink to `sparqlalgebrajs/node_modules/sparqljs`
8. verified `comunica/node_modules/sparqljs/package.json` to have ` "version": "3.2.0-beta.0"`

At this point I was reasonably sure that everything had been linked correctly. Run `yarn test` and everything passed. Running `yarn` _after_ linking `sparqljs` and `sparqlalgebrajs` still results in a bunch of TS-related errors (I get 28 of those) but they do look like strict vs. non-strict errors.

rubensworks:
Ok, so everything is working fine then.
I guess my problems were caused by multiple different instances of SPARQL.js.
Not sure if this is something you consider problematic @RubenVerborgh?

jacoscaz:
Just FYI, I just found out about Yarn's [selective dependency resolutions][1]. It would have made testing this a lot easier!

[1]: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

RubenVerborgh:
> I guess my problems were caused by multiple different instances of SPARQL.js.
> Not sure if this is something you consider problematic @RubenVerborgh?

I guess that is okay; it's essentially a bug in older SPARQL.js libraries that they are not compatible with newer versions of themselves 😅 Will release non-beta.

jacoscaz:
Thank you @RubenVerborgh ! @rubensworks @joachimvh , when will you be able to get this into SPARQLAlgebra.js and Comunica ?

On my end it’d be really nice to have a release of  SPARQLAlgebra.js incorporating this change as it would allow me to drop a few more kBs from our bundles right away and do a proper release (I don’t need to wait for Comunica due to how I am managing dependencies between quadstore and my comunica configuration).

Plus, Comunica just went through the 1.18.0 release and it probably wouldn’t make sense to do another one just for this.

rubensworks:
I can easily release a patch update for Comunica, that's not problem. But @joachimvh would have to update sparqlalgebrajs first then.

RubenVerborgh:
@jacoscaz You don’t depend on their updates, since this is semver minor. Just update all of your dependencies.

jacoscaz:
@RubenVerborgh You are totally correct and I should have realized that on my own. I tend to forget about the luxuries of semver...

Anyway, thank you ever so much and good weekend to you all!

PR #124, opened 2020-11-20
Add case insensitivity to absolute IRI detection

jeswr:
Part of resolving https://github.com/LDflex/LDflex-Comunica/issues/22 as blank nodes of the form `nodeID://1234` from Virtuoso need to be treated as absolute IRI's when going through this parser.


rubensworks (approved):
Not a must for me, but perhaps a unit test to show the impact might be valuable.

PR #132, opened 2021-02-16
Always update base to Parser.base

### LINKED ISSUE #130: Base IRI gets used in next request when parsing fails
Example code:
```typescript
const Parser = require('sparqljs').Parser;

try {
  const parser = new Parser({ baseIRI: 'http://localhost:3000/some/unrelated/resource.ttl' });
  parser.parse('DELETE DATA { <#something> <#completely> <#different> . } garbage')
} catch {}

const parser = new Parser({ baseIRI: 'http://localhost:3000/some/resource.ttl' });
const result = parser.parse('DELETE DATA { <#this> <#is> <#wrong> . }')
console.log(JSON.stringify(result, null, 2));
```
Outputs the following result:
```json
{
  "type": "update",
  "updates": [{
    "updateType": "delete",
    "delete": [{
      "type": "bgp",
      "triples": [{
        "subject": {
          "termType": "NamedNode",
          "value": "http://localhost:3000/some/unrelated/resource.ttl#this"
        },
        "predicate": {
          "termType": "NamedNode",
          "value": "http://localhost:3000/some/unrelated/resource.ttl#is"
        },
        "object": {
          "termType": "NamedNode",
          "value": "http://localhost:3000/some/unrelated/resource.ttl#wrong"
        }
      }]
    }]
  }],
  "base": "http://localhost:3000/some/resource.ttl",
  "prefixes": {}
}
```
So the `base` value is set to the correct value, but the generated URLs used the previous base. In case the first query is fixed so it doesn't error, this does not happen: the URLs in the second query will be correct.

I had a quick look into the code already, but not sure why this happens. My guess is that it might be related to https://github.com/RubenVerborgh/SPARQL.js/blob/e5bb5a6f062481ec68126d200a1625c3187b614a/lib/sparql.jison#L539 since that code only runs when a query gets parsed succesfully, so might be related to the internal base variables not being reset.

joachimvh:
Closes #130.

The internal `base` variable now always gets updated if it doesn't match `Parser.base`, which is what was already happening in case there was no crash.

Added relevant unit test that failed before this change.

(Also removed 2 useless test line).

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -61,7 +61,7 @@
       return iri;
     if (!Parser.base)
       throw new Error('Cannot resolve relative IRI ' + iri + ' because no base IRI was set.');
-    if (!base) {
+    if (base !== Parser.base) {
       base = Parser.base;
       basePath = base.replace(/[^\/:]*$/, '');
       baseRoot = base.match(/^(?:[a-z]+:\/*)?[^\/]*/)[0];
@@ -536,7 +536,7 @@ QueryOrUpdate
       $2 = $2 || {};
       if (Parser.base)
         $2.base = Parser.base;
-      Parser.base = base = basePath = baseRoot = '';
+      Parser.base = '';
       $2.prefixes = Parser.prefixes;
       Parser.prefixes = null;
       return $2;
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks!!

PR #133, opened 2021-02-17
Fix mocha setup

antoine-morvan:
- Use package.json config instead of deprecated opts file;
- Enable recursive test file lookup;

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,0 +1,5 @@
+{
+  "extension": ["js"],
+  "recursive": true,
+  "require": ["test/test-setup.js"]
+}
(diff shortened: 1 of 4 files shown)


RubenVerborgh (changes requested):
Thanks—let's not put it in package.json though; no need to publish this. What are the other options?

antoine-morvan:
Deprecation message states 2 options:
```
(node:30904) DeprecationWarning: Configuration via mocha.opts is DEPRECATED and will be removed from a future version of Mocha. Use RC files or package.json instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

I'll setup the RC file.

RubenVerborgh:
Thanks!!

PR #131, opened 2021-02-16
Added flag to disable checking of ungrouped variables

LaurensRietveld:
This PR should provide a workaround for the false errors described in #120

- I added a `skipUngroupedVariableCheck` option to skip checking for ungrouped variables. If you have a different name in mind, let me know (`strict` mode was already taken ;) )
- While I was at it, also added a test (set to `skip()`) that reproduces the issue from #120 in the test suite

PR #137, opened 2021-03-03
Fixed issue where we returned too early in a function call

LaurensRietveld:
I've made a mistake in PR #131 where I added a flag to skip ungrouped variables. In the previous implementation, we returned early from a function. Instead, I should have only skipped part of the function.

This PR includes the fix. I also updated the unit tests to avoid regressing in a similar fashion.


RubenVerborgh:
Thanks, merged as ba156dbb2be3f0a053fc4c944ca93a5871554ed2. Publishing new version.

Issue #99, opened 2020-06-25
Parsing issue with SPARQL query on inverse paths in anonymous node


rubensworks:
_Originally reported in by https://github.com/comunica/comunica/issues/677 @jaw111._

There is an odd combination of syntax that is flagged as an error.

```sparql
SELECT *
WHERE {
  ?movie a dbpedia-owl:Film ;
    dbpedia-owl:starring [
      rdfs:label "Brad Pitt"@en ;
      ^dbpedia-owl:starring ?otherMovie
    ] .
}
```

Gives the error:

```
# Parse error on line 21:
...ad Pitt"@en ;      ^dbpedia-owl:starrin
----------------------^
Expecting 'IRIREF', 'PNAME_NS', 'VAR', '}', 'GRAPH', ';', 'a', ']', 'PNAME_LN', '.', got '^'
```

Whilst removing the triple pattern for `a` works just fine:

```sparql
SELECT *
WHERE {
  ?movie
    dbpedia-owl:starring [
      rdfs:label "Brad Pitt"@en ;
      ^dbpedia-owl:starring ?otherMovie
    ] .
}
```

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Triply" src="https://nightly.triplydb.com/imgs/logos/logo-lg.svg?v=3" height="25">](https://triply.cc/) |
| €272 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

LaurensRietveld:
This probably isn't related to the type of property path operator, as this one fails as well:

```
prefix : <https://ex.com/>
select * {
:a :a :a ;
  :a [
    :a/:a :a
  ] .
}
```

wouterbeek:
Potentially useful, the following is the shortest query string that I can make for which this reproduces:

```sparql
ask { [<>?o]. }
```

The anonymous node notation is not necessary, I only use it for brevity.  The following also reproduces:

```sparql
ask { ?s<>?o. }
```

maxdebayser:
Hi Everyone, I've submitted a PR for this issue that appears to solve the query parsing problems reported in this issue, as well as some of my own queries. All pre-existing unit tests continue running correctly. 

Actually I wasn't able to reproduce the same parsing error with @wouterbeek 's queries, in both form, with or without my PR I get:
```
[...]

PR #142, opened 2021-04-23
Issue 99 - Fix blank node triple pattern with property paths

maxdebayser:
When blank node triple patterns of the form `[ :p "v" ] .` contained
property paths, the parsed throwed an exception because it didn't expect
it. This happened because the parser entered the ObjectList rule that
doesn't handle the property path case. So instead of having PropertyListPathNotEmptyTail
calling the ObjectList rule, I added a ObjectListPath rule that is
called instead and that calls GraphNodePath, following the same pattern
of -Path suffixed rules in this parser to make sure that property paths
arent allowed in other queries such as construct queries.

To the unit tests I've added the queries posted in the issue

https://github.com/RubenVerborgh/SPARQL.js/issues/99

Reference: https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#QSynBlankNodes

### DIFF
--- a/queries/sparql/sparql-18-b.sparql
+++ b/queries/sparql/sparql-18-b.sparql
@@ -1,0 +1,14 @@
+prefix example: <http://example/id/>
+PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
+PREFIX rdf:<http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+select ?s ?p ?o {
+    # ok
+    ?s  rdf:type [ example:prop* ?o ] .
+    # ok
+    ?s  rdf:type [ example:prop* ?o ] ;
+        example:p2 ?_ .
+    # error 
+    ?s  example:p1 ?_ ; 
+        rdf:type [ example:prop* ?o ] ;
+        example:p2 ?o .
+}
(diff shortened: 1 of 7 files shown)


queries/sparql/sparql-18-b.sparql
```diff
@@ -0,0 +1,14 @@
+prefix example: <http://example/id/>
+PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
+PREFIX rdf:<http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+select ?s ?p ?o {
+    # ok
+    ?s  rdf:type [ example:prop* ?o ] .
+    # ok
+    ?s  rdf:type [ example:prop* ?o ] ;
+        example:p2 ?_ .
+    # error
```
RubenVerborgh, line 10: Should these comments be removed?

RubenVerborgh (approved):
Looks good, thanks. Just trying to understand the comments; they can go now?

queries/sparql/sparql-18-b.sparql
maxdebayser: Sorry, I forgot to clean that up. It's fixed now.

RubenVerborgh:
Thank you!

Issue #120, opened 2020-10-29
Throwing `Projection of ungrouped variable` error on valid query


LaurensRietveld:
We're encountering issues with the most recent sparqljs releases, where the error checking is too strict.
This query is valid, but throws an `Projection of ungrouped variable (?p)` error:

```
select ?p (count(*) as ?n) {
  [ ?p [] ].
}
```

This seems to be introduced by this PR: #108

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Triply" src="https://nightly.triplydb.com/imgs/logos/logo-lg.svg?v=3" height="25">](https://triply.cc/) |
| €272 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

LaurensRietveld:
@RubenVerborgh This still seems to be an issue. These two queries fail using the latest version:

RubenVerborgh:
@rubensworks I might need to revert the protection mechanisms that have been built in; it seems that they are not created adequately. This probably will have repercussions on our spec compliance?

rubensworks:
Instead of reverting, perhaps we should instead look into fixing the problem(s) with the protection mechanisms?
It looks to me like reverting would break more things than it would fix.

An alternative would be to add an option to disable the protection mechanisms, which could be used until all (known) issues with it are fixed.

RubenVerborgh:
> Instead of reverting, perhaps we should instead look into fixing the problem(s) with the protection mechanisms?

I will try; it's just the second issue in a very short time. I'm not sure whether they were properly designed.

wouterbeek:
I'm not sure whether this is the same error or a related by different one: when a `group by` clause is used to explicitly specify the grouped-by variables, and a variable appears in the projection but not in an aggregate function, the variable should be sampled (i.e., the equivalent of applying the aggregate function `sample` explicitly).

This means that the following query string should parse, but at the moment it does not:

```sparql
select ?s (?x as ?y) {
  ?s ?p ?x.
}
group by ?s
```

^ The above query string is parsed by Virtuoso but is not parsed by Jena :-/

The relevant section in the standard is <https://www.w3.org/TR/sparql11-query/#sparqlGroupAggregate>.

LaurensRietveld:
[...]

PR #145, opened 2021-09-06
Implicit grouping by count throws 120

danielbeeke:
Hi Ruben and others,

I think I may have finished #120.

- Renamed skipUngroupedVariableCheck to skipValidation
- Added skipValidation to readme
- I did not find other places where adding skipValidation looked useful. If this is incorrect I can add it. 
- Added a new folder for testing queries that only work while skipping the validation
- Added one query that should have been added before. It was at the top of #120 
- Fixed the call to .map when there was no array
- Spec seems to work, tests also. It does however give notices about .tsv files.


danielbeeke:
One small thing that I am wondering about.. the names of the queries do they have a relationship with the spec?
If so I might need to rename the newly added queries.

README.md
```diff
@@ -28,6 +28,24 @@ parsedQuery.variables = ['?mickey'];
 var generatedQuery = generator.stringify(parsedQuery);
 ```
 Set `sparqlStar` to `true` to allow [SPARQL*](https://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/) syntax.
+
+
+### Validation
+
```
RubenVerborgh, line 34:
```suggestion

### Validation
```

README.md
```diff
@@ -28,6 +28,24 @@ parsedQuery.variables = ['?mickey'];
...
+By default SPARQL.js throws on validation errors.
```
RubenVerborgh, line 35: Perhaps explain what they are; "query validation" as such is not a thing.

Maybe queries that are syntactically correct, but not allowed by the spec?

README.md
```diff
@@ -28,6 +28,24 @@ parsedQuery.variables = ['?mickey'];
...
+
+```JavaScript
+// Parse a SPARQL query without validation.
+var SparqlParser = require('sparqljs').Parser;
+var parser = new SparqlParser({ skipValidation: true });
+var parsedQuery = parser.parse(
+  'select (?x as ?xString)' +
+  '(count(?y) as ?count)' +
+  '{ ?x ?y ?z }');
+
```
RubenVerborgh, line 46:
```suggestion
  '{ ?x ?y ?z }');
```

(remove the newline)

lib/sparql.jison
```diff
@@ -642,7 +642,7 @@ SelectQuery
     | SelectClauseVars     DatasetClause* WhereClause SolutionModifier
     {
       // Check for projection of ungrouped variable
-      if (!Parser.skipUngroupedVariableCheck) {
+      if (!Parser.skipValidation) {
```
RubenVerborgh, line 645:
```suggestion
      if (!Parser.skipUngroupedVariableCheck && !Parser.skipValidation) {
```

so we maintain compatibility for the entire major version.

sparql.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
    *   sparqlStar?: boolean,
    * }
    */
-  Parser: function ({ prefixes, baseIRI, factory, sparqlStar, skipUngroupedVariableCheck } = {}) {
+  Parser: function ({ prefixes, baseIRI, factory, sparqlStar, skipValidation } = {}) {
```
RubenVerborgh, line 16: We need to keep it here too for backward compatibility.

RubenVerborgh:
>  the names of the queries do they have a relationship with the spec?

Yes, queries like `sparql-10-2-2c.sparql` point to section 10.2.2 of the spec.

So probably best to rename indeed.

RubenVerborgh (changes requested):
Thanks, @danielbeeke. Looks good, except that we need to maintain backwards compatibility, so the existing option flag still has to work. I'm fine with introducing another name as well; just both have to work.

rubensworks:
Thanks @danielbeeke!

> Spec seems to work, tests also. It does however give notices about .tsv files.

You can ignore this, it's a warning that is unrelated to SPARQL.js.

lib/sparql.jison
danielbeeke: I fixed it in sparql.js, that way the parser code looks cleaner to me. Would this be a good way or would you rather have it in the way you have suggested?

RubenVerborgh:
All good, thanks!

PR #146, opened 2021-09-06
Adding query of issue #99 so we have a regression this for this use case

danielbeeke:
Hi Ruben and others,

As discussed I looked into #99. It looks like there are no problems with that query at the moment.
Some other commit may have fixed it.

I added the query so we have a regression test.


RubenVerborgh:
Super, thanks @danielbeeke!

PR #148, opened 2021-10-23
Inverse paths in anonymous node 99

danielbeeke:
I have added the regression tests.
Number 4 seems to be invalid, I added that one also in case it will ever turn out to be valid while a developer works in the code.

I propose the following naming scheme for regression queries:

regression-github-ISSUE_NUMBER-occurance-of-query

Is there an existing naming scheme or do you think there would a better alternative?


danielbeeke:
I have changed the names, do you want to check if the names make sense?

rubensworks (approved):
@danielbeeke Looks good to me, thanks!

@RubenVerborgh This should now fully resolve #99.

RubenVerborgh:
Thank you!

Issue #150, opened 2022-04-12
Incorrect serialization of ADD, MOVE, COPY statements in SPARQL generator.


Ladone3:
Hi,
at metaphacts we found that the SPARQL generator does omit the source graph when serializing SPARQL 1.1. Update `ADD`, `MOVE` or `COPY` operations (c.f. https://www.w3.org/TR/sparql11-update/#copy) .
For example, ```ADD <http://example.com/my-graph> TO <http://example.com/your-graph>```  becomes ```ADD TO <http://example.com/your-graph>``` after the parse/serialize roundtrip (c.f. https://github.com/metaphacts/SPARQL.js/commit/05a334d5f83bb7d2016d92bc9c0e892c90d937a9 for failing test). We will bring up a PR in a minute.

Issue #100, opened 2020-07-01
Inconsistencies in update queries


joachimvh:
I have found some inconsistencies in the output of some update queries.

1. Some updates have a [type](https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/parsedQueries/sparql-update-3-2-4.json#L5), while others have an [updateType](https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/parsedQueries/sparql-update-1-1-1.json#L5) instead.
2. In case of the default graph (or any other reference to graphs that is not an IRI), sometimes it has [type graph](https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/parsedQueries/sparql-update-3-2-5.json#L7-L10), and sometimes it [doesn't](https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/parsedQueries/clear-drop.json#L51-L53).
3. This one is more debatable, but graphs are represented differently than in SELECT queries. For SELECT queries, you have a [graph statement](https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/parsedQueries/sparql-13-3-4.json#L52-L92) with a nested BGP statement. For updates, the [graph statement](https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/parsedQueries/sparql-update-3-1-1b.json#L8-L34) immediately contains the triples. In the first case the graph can contain any possible SPARQL statement and in the second case only triple patterns though, so I can see why this one might be done this way.

PR #151, opened 2022-04-12
Fix sparql generator for SPARQL 1.1 update ADD, MOVE, COPY operations

QuarksToQuasars:
Fixing #150.
What is a bit strange IMHO, is that in the standard the GRAPH keyword seems to be optional for ADD, MOVE, COPY (c.f.  https://www.w3.org/TR/sparql11-update/#COPY and https://www.w3.org/TR/sparql11-query/#rGraphOrDefault , ) but not for others e.g. https://www.w3.org/TR/sparql11-update/#DROP ("The GRAPH keyword is used to remove a graph denoted by IRIref, the DEFAULT keyword is used to remove the default graph from the Graph Store").

 So I haven't checked yet whether it is possible to preserve the GRAPH, for example, `ADD GRAPH <http://example.com/my-graph> TO GRAPH <http://example.com/your-graph>` instead of `ADD <http://example.com/my-graph> TO <http://example.com/your-graph>` . In the current fix it would simply drop the GRAPH keyword in the roundtrip. I guess that may relate to https://github.com/RubenVerborgh/SPARQL.js/issues/100 ?

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -382,8 +382,8 @@ Generator.prototype.toUpdate = function (update) {
   case 'add':
   case 'copy':
   case 'move':
-    return update.type.toUpperCase() + (update.source.default ? ' DEFAULT ' : ' ') +
-           'TO ' + this.toEntity(update.destination.name);
+    return update.type.toUpperCase()+ ' ' +  (update.silent ? 'SILENT ' : '') + (update.source.default ? 'DEFAULT' : this.toEntity(update.source.name)) +
+           ' TO ' + this.toEntity(update.destination.name);
   case 'create':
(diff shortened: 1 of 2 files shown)


QuarksToQuasars:
Hi @RubenVerborgh . Any thoughts on this?

RubenVerborgh:
Catching up after some health issues, I'll be back soon!

Ladone3:
Hello, did you have a chance to look at this PR?

RubenVerborgh:
Catching up after some health issues, I'll be back soon!

RubenVerborgh (approved):

RubenVerborgh:
Super, thank you!

PR #155, opened 2022-06-13
Feature/path only parsing

jeswr:
This PR adds the feature to parse only SPARQL paths rather than the full SPARQL algebra.

This is of use in several downstream applications including [here](https://github.com/LDflex/LDflex/blob/9b188bb35a88a5fd085de270900afce8e8a04e9d/src/ComplexPathResolver.js#L70-L74) in LDflex


RubenVerborgh:
I like the idea, but do we have other usage besides LDflex?

lib/sparql.jison
```diff
@@ -583,6 +583,15 @@ QueryOrUpdate
       $2.prefixes = Parser.prefixes;
       Parser.prefixes = null;
 
+      if (Parser.pathOnly) {
+        if ($2.type === 'path' || 'termType' in $2) {
+          return $2
+        }
+        throw new Error('Recieved full SPARQL query in path only mode');
```
RubenVerborgh, line 590:
```suggestion
        throw new Error('Received full SPARQL query in path only mode');
```

lib/sparql.jison
```diff
@@ -583,6 +583,15 @@ QueryOrUpdate
...
+      } else if ($2.type === 'path' || 'termType' in $2) {
+        throw new Error('Recieved only path in full SPARQL mode');
```
RubenVerborgh, line 592:
```suggestion
        throw new Error('Received only path in full SPARQL mode');
```

README.md
```diff
@@ -29,6 +29,8 @@ var generatedQuery = generator.stringify(parsedQuery);
 ```
 Set `sparqlStar` to `true` to allow [SPARQL*](https://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/) syntax.
 
+Set `pathOnly` to `true` to parse SPARQL paths such as `foaf:name/foaf:knows` rather than the full SPARQL Algebra.
```
RubenVerborgh, line 32: Note that the `pathOnly` option does not enforce only paths; so perhaps better `acceptPaths` or so?

RubenVerborgh: Or of course the reverse, where it MUST be a path.

jeswr:
> it seems that LDflex could just slap a SELECT * around it?

Correct that's the current behavior

> other usage

Not that I have top of mind,  but I wouldn't be surprised if I need the same behavior in other libs down the line. Happy to park it until that occurs.

README.md
jeswr: > Note that the pathOnly option does not enforce only paths.

That's exactly what the second error is for - I just don't have good test coverage for that part yet

RubenVerborgh:
> Happy to park it until that occurs.

We can do it; let's just check the final bundle size. Jison can be weird like that.

README.md
RubenVerborgh: Makes sense!

jeswr:
> final bundle size

The generated `SparqlParser.js` increases by ~0.5% (112K before and after if you round).

RubenVerborgh:
Perfect, that's okay with me.

Are you happy with coverage etc.?
Maybe just rename `paths/al` folder into `queries/paths`, and similar for `test`.

jeswr:
>  Maybe just rename paths/al folder into queries/paths, and similar for test.

Done

> Are you happy with coverage etc.?

I am now after adding those tests for erroring :)

jeswr:
/bump @RubenVerborgh - this should be ready to merge

PR #158, opened 2022-08-28
Chore/update

### LINKED ISSUE #157: Spec tests broken on Node 18
Errors with `body.pipe` is not a function. Likely caused by introduction of native fetch in Node 18.

jeswr:
resolves #157 

We can no longer support Node 8 and 10 as they do not have `globalThis`; which is used by the nested `readable-stream` dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,6 @@ jobs:
     strategy:
       matrix:
         node-version:
-          -  8.x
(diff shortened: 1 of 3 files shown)


jeswr:
bump @RubenVerborgh

RubenVerborgh:
Thanks, Jesse. That's unfortunate, but we need to move on indeed.

Issue #139, opened 2021-03-23
Remove built-ins from SparqlParser.js


tpluscode:
The module `lib/SparqlParser.js` wants to `require` the node modules `fs` and `path`

This is a bit problematic in my setup where I want run browser tests for a component which uses the parser. It is running rollup and, while commonjs can be easily transformed, I'm having a hard time setting it up so that it doesn't choke on the built-in modules.

Would it be possible to move the function `exports.main = function commonjsMain (args) {}` somewhere else?

RubenVerborgh:
Mmm, the prob is that `lib/SparqlParser.js` is autogenerated.

tpluscode:
Thought so :(. I did a dirty, hacky rewrite in the test setup for now...

jacoscaz:
I suspect that this is also what is preventing usage of `quadstore-comunica` in Deno:

```sh
error: Uncaught ReferenceError: require is not defined
    if (require.main === module) {
    ^
    at https://cdn.skypack.dev/-/sparqljs@v3.5.2-dsMDqK77bLuGqQk32ifA/dist=es2019,mode=imports/optimized/sparqljs.js:1900:5
    at createCommonjsModule (https://cdn.skypack.dev/-/sparqljs@v3.5.2-dsMDqK77bLuGqQk32ifA/dist=es2019,mode=imports/optimized/sparqljs.js:204:6)
    at https://cdn.skypack.dev/-/sparqljs@v3.5.2-dsMDqK77bLuGqQk32ifA/dist=es2019,mode=imports/optimized/sparqljs.js:239:22
```

Otherwise, Skypack's version available at https://cdn.skypack.dev/sparqljs@v3.5.2 should be good to go. I can't think of anything outside super-hacky solutions... @tpluscode how have you addressed this?

tpluscode:
@jacoscaz this is super specific to [@web/test-runner](https://www.npmjs.com/package/@web/test-runner) https://github.com/zazuko/rdfjs-elements/blob/27c7b29274077a6f6e281d94fcf747b06bb55ddf/web-test-runner.config.mjs#L23-L32

jacoscaz:
@RubenVerborgh would you accept a PR that manually removes that code? It's bound to be at least somewhat hacky, though.

RubenVerborgh:
This whole library has always been a hack in the first place (was supposed to be a parser for 1 summer until something better was written). So yeah, with a repeatable build process, why not!

[...]

PR #159, opened 2022-09-28
Adds script to drop require() invocation to native modules at build time

jacoscaz:
Hack-ish solution to solve #139 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "bin/*"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
@jacoscaz Thanks! How do you like 055cae12c99a14bcc25d69f00846ca13035ade93?

jacoscaz:
@RubenVerborgh that's indeed *a lot* better than my solution. Still a bit hacky, but a lot less hacky.

RubenVerborgh:
Wouldn't have found this without your start on this, so great!

PR #171, opened 2023-07-14
Make exports compatible with ES modules

Peeja:
Node attempts to make CommonJS exports available as named exports when imported by ES modules. It does this by statically analyzing the CJS module using `cjs-module-lexer`. However, as the name implies, it's a lexer, not a full parser, and it only supports limited common patterns for building a `module.exports` object. Notably, it supports `module.exports = { ... }`, but only if the values in that object are all simple identifiers. Otherwise it will include only the keys up to the first non-identifier value it finds, and bail after that.

https://github.com/nodejs/cjs-module-lexer/issues/47

That means that the `Generator` and `Wildcard` exports were missing in ES modules. This change makes them available again.

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -447,7 +447,7 @@ function mapJoin(array, sep, func, self) {
  *   newline: string
  * }
  */
-module.exports = function SparqlGenerator(options = {}) {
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks @Peeja, good point indeed.

PR #177, opened 2024-08-09
Expand regex for absolute IRIs

### LINKED ISSUE #176: URI schemes containing digits cause parse error
I have some RDF data autoconverted from a Neo4j database. It creates URIs using a `neo4j://` scheme like this:

```turtle
<neo4j://graph.individuals#n4ind:389638>  <neo4j://graph.schema#knows>  <neo4j://graph.individuals#145613> .
```

Using this data with Comunica throws an error in SPARQL query parsing:

`Cannot resolve relative IRI neo4j://graph.individuals#105029 because no base IRI was set.`

It looks like this is caused by the regex here: https://github.com/RubenVerborgh/SPARQL.js/blob/ebde86c4bbc52ee1356f10f9c539a30441c8af1d/lib/sparql.jison#L60

which only allows letters. I haven't found anything that says a scheme can't contain a digit, and [Wikipedia](https://en.wikipedia.org/wiki/List_of_URI_schemes#Unofficial_but_common_URI_schemes) also mentions `s3://` as a commonly used scheme.

balhoff:
Fixes #176.

I updated to the regex to match the grammar in https://www.ietf.org/rfc/rfc2396.txt:

>Scheme names consist of a sequence of characters beginning with a
   lower case letter and followed by any combination of lower case
   letters, digits, plus ("+"), period ("."), or hyphen ("-").  For
   resiliency, programs interpreting URI should treat upper case letters
   as equivalent to lower case in scheme names (e.g., allow "HTTP" as
   well as "http").
>
>      scheme        = alpha *( alpha | digit | "+" | "-" | "." )

### DIFF
--- a/lib/sparql.jison
+++ b/lib/sparql.jison
@@ -57,7 +57,7 @@
     if (iri[0] === '<')
       iri = iri.substring(1, iri.length - 1);
     // Return absolute IRIs unmodified
-    if (/^[a-z]+:/i.test(iri))
+    if (/^[a-z][a-z0-9.+-]*:/i.test(iri))
       return iri;
     if (!Parser.base)
       throw new Error('Cannot resolve relative IRI ' + iri + ' because no base IRI was set.');


RubenVerborgh:
Thanks @balhoff, good catch! Your code is correct; I just slightly changed the order from `a-z0-9.-+` to `a-z0-9.+-` so readers don't confuse it as "a to z, 0 to 0, dot to plus".

balhoff:
> Thanks @balhoff, good catch! Your code is correct; I just slightly changed the order from `a-z0-9.-+` to `a-z0-9.+-` so readers don't confuse it as "a to z, 0 to 9, dot to plus".

Good idea, thank you!

PR #184, opened 2024-11-20
Preserve LIMIT 0

### LINKED ISSUE #183: LIMIT 0 should be preserved
https://www.w3.org/TR/sparql11-query/#modResultLimit

> A LIMIT of 0 would cause no results to be returned. A limit may not be negative.

```it('should preserve limits with 0', function () {
    var parser = new SparqlParser();
    var parsedQuery = parser.parse('SELECT * WHERE { ?s ?p ?o. } LIMIT 0');
    var generator = new SparqlGenerator();
    var generatedQuery = generator.stringify(parsedQuery);
    var expectedQuery =
      'SELECT * WHERE { ?s ?p ?o. }\n' +
      'LIMIT 0';
    expect(generatedQuery).toEqual(expectedQuery);
  });
  ```
  
  fails as the generator omits the limit in case of 0 (https://github.com/RubenVerborgh/SPARQL.js/blob/main/lib/SparqlGenerator.js#L79)

QuarksToQuasars:
LIMIT 0 should not be dropped / database should return 0 results. 

See https://www.w3.org/TR/sparql11-query/#modResultLimit

Fixes #183

### DIFF
--- a/lib/SparqlGenerator.js
+++ b/lib/SparqlGenerator.js
@@ -76,7 +76,8 @@ Generator.prototype.toQuery = function (q) {
 
   if (q.offset)
     query += 'OFFSET ' + q.offset + this._newline;
-  if (q.limit)
+  //do integer check to preserve LIMIT 0
+  if (Number.isInteger(q.limit))
     query += 'LIMIT ' + q.limit + this._newline;
 
   if (q.values)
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks, merged as 13cc3d2ee4d2528b85d8b25cfbf886597dd100c1!

PR #188, opened 2025-07-28
Update rdf-data-factory to v2.

jahed:
This PR also includes a fix to serialise Wildcards to JSON, which was needed to correctly update test snapshots. Without the fix, Wildcards serialise to empty object (`{}`).

I noticed a bunch of other dependencies are on v1, so leaving this as draft.

### DIFF
--- a/lib/Wildcard.js
+++ b/lib/Wildcard.js
@@ -2,24 +2,27 @@
 // Wildcard constructor
 class Wildcard {
   constructor() {
-    return WILDCARD || this;
+    if (WILDCARD) {
+      return WILDCARD;
+    }
+
+    Object.defineProperty(this, 'termType', {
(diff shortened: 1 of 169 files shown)


jahed:
Updating `rdf-test-suite` and in turn `jsonld-streaming-parser` to use `rdf-data-factory@2` requires a lot of work to detangle. The latest `rdf-string` isn't parsing `--direction` properly, and so on. These are all dev dependencies which affect tests. sparqljs itself works fine with v2. So in the mean time it's safe to use `overrides` in `package.json` instead of waiting for sparqljs to update.

```json
{
  "overrides": {
    "sparqljs": {
      "rdf-data-factory": "^2"
    }
  }
}
```

